PR #185, opened 2020-02-01
Member subscription management

Floury-Fabio:


.gitignore
```diff
@@ -29,3 +29,6 @@ coverage
 # Ignore master key for decrypting credentials and more.
 /config/master.key
 .env
+
+*.swp
```
Mate2xo: Since those files are generated on your environment only (I guess it's done by your VIM text editor), putting them here would not be relevant for other developers. You can follow [this procedure](https://help.github.com/en/github/using-git/ignoring-files#explicit-repository-excludes) to exclude them from your commits locally, and remove them from this `.gitignore` file

app/mailers/member_mailer.rb
```diff
@@ -0,0 +1,11 @@
+class MemberMailer < ApplicationMailer
+
+	default from: 'no-reply@monsite.fr'
```
Mate2xo: If you want to set a dummy default, please edit the one from `application_mailer.rb`
e.g. : ``` default from: Rails.application.credentials.dig(:noreply_mail_address) || 'no-reply@example.com' ```

app/mailers/member_mailer.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  def alert_end_of_souscription_email(member)
+    @member = member
```
Mate2xo, line 8: Do you need to access `member` as an instance variable somewhere else in the code? If not, you can use the `member` variable directly (e.g. ```member.email```), and delete the line `@member = member`

app/models/member.rb
```diff
@@ -72,6 +73,34 @@ def thredded_admin?
     admin? || super_admin?
   end
 
+	def renewed_subscription
```
Mate2xo: If I understand correctly, this method is used to renew a subscription date.
In that case, it should be renamed as `renew_subscription_date` ('renewed_subsciption' may convey that this method is used to fetch the last renewed subscription).

Also, where do you intend to use this method? I don't see it anywhere outside `member_spec.rb`

app/services/end_souscription_notification.rb
```diff
@@ -0,0 +1,13 @@
+class EndSouscriptionNotification 
+	def perform
+
+		n = Time.now
```
Mate2xo: `n` variable should be rename to `now`. Please avoid single character variables.

app/services/end_souscription_notification.rb
```diff
@@ -0,0 +1,13 @@
...
+		start = Time.new(n.year, n.month, n.day, 23, 30)
+
+		schedule = IceCube::Schedule.new(start)
+		schedule.add_recurrence_rule IceCube::Rule.daily
+		singleton.each do
```
Mate2xo: What is `singleton` ? I guess that the intent of this service is to launch `SendNotificationJob` on a daily basis? I don't see it called anywhere else in the code.

app/views/member_mailer/alert_end_of_souscription_email.html.erb
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <h1>Bonjour <%= @member.first_name @member.last_name%></h1>
```
Mate2xo: Please use the `I18n` API. You can create or edit locale files in `/config/locale/` folder

app/views/member_mailer/alert_end_of_souscription_email.txt.erb
```diff
@@ -0,0 +1,8 @@
+Bonjour <%= @member.first_name @member.last_name%>
+==========================================================
+ 
+ 
+Tu t'es inscrit sur monsite.fr en utilisant l'e-mail suivant : <%= @user.email %>.
```
Mate2xo: This message should be same as in the HTML version.
You can use the `I18n` API here too.

app/views/members/_form.html.erb
```diff
@@ -57,6 +57,15 @@
           {},
           class:'form-control' %>
       </div>
+      <br>
+      <div class="field">
+				<%= form.label :end_subscription, Member.human_attribute_name(:votre_adhésion_prend_fin_le).humanize %>
+			</br>
```
Mate2xo, line 64: `#human_attribute_name` is from the `I18n` API, and is used to translate a Model's attribute to a human-readable format.
So in this case, you should add a key in `/config/locale/fr.yml` under `"activerecord.attributes.member"`, and give the `:end_subscription` attribute as the first argument to `#human_attribute_name`. The `I18n` API will then fetch the localized string associated with that attribute.

This `form.label` line should be in the following `if ... else` clause, as a lonely label without it's form would be a confusing UX

app/views/members/show.html.erb
```diff
@@ -62,6 +63,21 @@
...
+
+						<hr class="mt-0 mb-0">
+
+            <p class="text-lg-7 line-height-1_8 mt-4">
+							<% if @member.end_subscription.present? %>
+								<%= @member.end_subscription.strftime('%d-%m-%Y') %>
+							<% end %>
+						</p>
+					</div>
+
```
Mate2xo: This information should be private, and must not be shown in `members/show.rb`.
Only admins, in the admin interface, should be able to see this information about other members.
You can delete this section

spec/models/member_spec.rb
```diff
@@ -160,4 +163,165 @@
       it { is_expected.not_to be_thredded_admin }
     end
   end
+
+	describe "renewed_subscription" do
+
+			before(:each) do
+				@member = FactoryBot.create(:member)
+			end
```
Mate2xo: You can replace this `before` setup with `let(:member) { create :member }`
This will scope the `member` variable to the current example block, instead of creating a `@member` instance variable that can be accessed from anywhere in the class.

FactoryBot is also included directly in `spec_helper.rb`, so you can omit `FactoryBot.` when calling FactoryBot methods

spec/models/member_spec.rb
```diff
@@ -160,4 +163,165 @@
...
+
+		context "when is the first subscription" do
```
Mate2xo: Do you mean "when a member subscribes for the first time" ?
In that case, it could be useful to show this in the test setup, in that example group

spec/models/member_spec.rb
```diff
@@ -160,4 +163,165 @@
...
+			it { 
+				allow(Date).to receive(:today) { Date.new(2060, 2, 29) }
+				@member.renewed_subscription
+				expect(@member.end_subscription).to eq(Date.new(2061, 2, 28)) 
+			}
+
+		end
+
+		context "when the subscription is outdated" do
+
```
Mate2xo: How does the test know that the subscription is outdated? You may need to set it up for this example group.

spec/views/members/edit.html.erb_spec.rb
```diff
@@ -9,4 +9,15 @@
...
+RSpec.describe "members/show.html.erb", type: :view do
+	context 'it display the end_subscription' do
+		it "displays end_subscription" do
+			assign(:member, build(:member, end_subscription: Date.new(2020, 10, 10) ))
+      allow(view).to receive(:current_member) { build_stubbed :member }
+			render
+			expect(rendered).to match match '10-10-2020' 
+		end
+	end
+end
```
Mate2xo: This is a duplicate from `members/show.html.erb_spec.rb` and should be deleted

spec/views/members/show.html.erb_spec.rb
```diff
@@ -12,4 +12,15 @@
...
+RSpec.describe "members/show.html.erb", type: :view do
+	context 'it display the end_subscription' do
+		it "displays end_subscription" do
+			assign(:member, build(:member, end_subscription: Date.new(2020, 10, 10) ))
+      allow(view).to receive(:current_member) { build_stubbed :member }
+			render
+			expect(rendered).to match '10-10-2020' 
+		end
+	end
+end
```
Mate2xo: This information should not appear on members/show.html, but should instead appear in the admin interface.

spec/models/member_spec.rb
```diff
@@ -160,4 +163,165 @@
...
+				allow(Date).to receive(:today) { Date.new(2399, 10, 24) }
+				@member.renewed_subscription
+				expect(@member.end_subscription).to eq(Date.new(2400, 10, 23)) 
+			}
+
+			it { 
+				allow(Date).to receive(:today) { Date.new(2060, 2, 29) }
+				@member.renewed_subscription
+				expect(@member.end_subscription).to eq(Date.new(2061, 2, 28)) 
+			}
```
Mate2xo: You are testing the same single case multiple times : checking that the new subscription date will equal `today + 1 year`
If you want to test random values, you can set `today` as a random date.

app/mailers/member_mailer.rb
```diff
@@ -0,0 +1,11 @@
+class MemberMailer < ApplicationMailer
+
+	default from: 'no-reply@monsite.fr'
+
+  def alert_end_of_souscription_email(member)
```
Mate2xo: `souscription` -> `subscription`

app/jobs/send_notification_job.rb
```diff
@@ -0,0 +1,15 @@
+class SendNotificationJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+
+		members = Member.all
+
+		members.each do |member|
+			if member.end_subscription.present? && ((member.end_subscription - Date.today == 7) || (member.end_subscription - Date.today == 3))
+				MemberMailer.welcome_email(member).deliver_now
```
Mate2xo: I guess you meant `MemberMailer.alert_end_of_subcription` ?

Mate2xo (changes requested):
Thank you for your work! We should meet up to set up the mailer notification logic. In the meantime, I hope my comments will help you.
Also, the `rubocop` command is available to you to format your code to the app's format.

PR #237, opened 2020-11-05
Time accounting

Floury-Fabio:
https://github.com/Mate2xo/Creeons_la_coop/issues/229

### DIFF
--- a/app/admin/members.rb
+++ b/app/admin/members.rb
@@ -18,16 +18,18 @@ ActiveAdmin.register Member do
                 :register_id,
                 group_ids: []
 
+  decorate_with MemberDecorator
+
   index do
     selectable_column
     column :first_name
     column :last_name
     column :role
-    column('3 heures faites?') { |member| member.worked_three_hours?(Date.current) }
     column(:group) do |member|
       group_links = member.groups.map { |group| auto_link group }
       safe_join group_links, ', '
     end
+    column :worked_hours_of_three_last_months
     column :cash_register_proficiency
     column :register_id
     column :email
@@ -40,10 +42,8 @@ ActiveAdmin.register Member do
     column :last_name
     column :phone_number
     column :role
-    column("3 heures de #{l 1.month.ago, format: '%B'}?") { |member| member.worked_three_hours?(1.month.ago) }
-    column("3 heures de #{l Date.current, format: '%B'}?") { |member| member.worked_three_hours?(Date.current) }
-    column("3 heures de #{l 1.month.from_now, format: '%B'}?") { |member| member.worked_three_hours?(1.month.from_now) }
     column(:group) { |member| member.groups.map(&:name).join(', ') }
+    column :worked_hours_of_three_last_months_csv_version
     column :cash_register_proficiency
     column :register_id
   end
--- a/app/admin/members.rb
+++ b/app/admin/members.rb
@@ -29,7 +29,7 @@ ActiveAdmin.register Member do
       group_links = member.groups.map { |group| auto_link group }
       safe_join group_links, ', '
     end
-    column(t('.worked_hours')) { |member| MemberDecorator.new(member).hours_worked_in_the_three_last_months }
+    column(t('.worked_hours')) { |member| member.hours_worked_in_the_three_last_months }
     column :cash_register_proficiency
     column :register_id
     column :email
@@ -43,7 +43,7 @@ ActiveAdmin.register Member do
     column :phone_number
     column :role
     column(:group) { |member| member.groups.map(&:name).join(', ') }
-    column(t('active_admin.resource.index.worked_hours')) { |member| MemberDecorator.new(member).hours_worked_in_the_three_last_months(csv: true) }
+    column(t('active_admin.resource.index.worked_hours')) { |member| member.hours_worked_in_the_three_last_months(csv: true) }
     column :cash_register_proficiency
     column :register_id
   end
--- a/app/admin/members.rb
+++ b/app/admin/members.rb
@@ -29,7 +29,7 @@ ActiveAdmin.register Member do
       group_links = member.groups.map { |group| auto_link group }
       safe_join group_links, ', '
     end
-    column(t('.worked_hours')) { |member| member.hours_worked_in_the_three_last_months }
+    column(t('.worked_hours')) { |member| member.hours_worked_in_the_last_three_months }
     column :cash_register_proficiency
     column :register_id
     column :email
@@ -43,7 +43,7 @@ ActiveAdmin.register Member do
     column :phone_number
     column :role
     column(:group) { |member| member.groups.map(&:name).join(', ') }
-    column(t('active_admin.resource.index.worked_hours')) { |member| member.hours_worked_in_the_three_last_months(csv: true) }
+    column(t('active_admin.resource.index.worked_hours')) { |member| member.hours_worked_in_the_last_three_months(csv: true) }
     column :cash_register_proficiency
     column :register_id
   end
--- a/app/admin/members.rb
+++ b/app/admin/members.rb
@@ -29,7 +29,7 @@ ActiveAdmin.register Member do
       group_links = member.groups.map { |group| auto_link group }
       safe_join group_links, ', '
     end
-    column :worked_hours_of_three_last_months
+    column(t('.worked_hours')) { |member| MemberDecorator.new(member).hours_worked_in_the_three_last_months }
     column :cash_register_proficiency
     column :register_id
     column :email
@@ -43,7 +43,7 @@ ActiveAdmin.register Member do
     column :phone_number
     column :role
     column(:group) { |member| member.groups.map(&:name).join(', ') }
-    column :worked_hours_of_three_last_months_csv_version
+    column(t('active_admin.resource.index.worked_hours')) { |member| MemberDecorator.new(member).hours_worked_in_the_three_last_months(csv: true) }
     column :cash_register_proficiency
     column :register_id
   end
--- a/app/decorators/member_decorator.rb
+++ b/app/decorators/member_decorator.rb
@@ -1,21 +1,24 @@
+# frozen_string_literal: true
+
+# MemberDecorator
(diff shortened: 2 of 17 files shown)


app/decorators/member_decorator.rb
```diff
@@ -0,0 +1,21 @@
+class MemberDecorator < Draper::Decorator
+  delegate_all
+
+  def worked_hours_of_three_last_months
```
Mate2xo: -> `hours_worked_in_the_last_three_months`

app/decorators/member_decorator.rb
```diff
@@ -0,0 +1,21 @@
...
+    hours_per_month = []
+    3.times do |n|
+      hours_per_month[n] = h.content_tag(:p, "#{I18n.localize(Date.current - n.month, format: :only_month)} :
+                                              #{model.monthly_worked_hours(Date.current - n.month)}")
+    end
+    h.safe_join hours_per_month.reverse
+  end
+
+  def worked_hours_of_three_last_months_csv_version
```
Mate2xo: Since those two methods are nearly the same, you could refactor to combine them together, by giving a `csv: true` keyword argument to signify you want the csv version. E.g.: `def hours_worked_in_the_last_three_months(csv: true)`
You could then separate the steps according to that option.
Also, instead of declaring `hours_per_month = []` before the `each` loop, you can try the ruby method `.with_object`

For example:
```ruby
def hours_worked_in_last_three_months(csv: false)
  hours_per_month = 3.times.with_object([]) do |n, array|
                      array[n] = "#{I18n.localize(Date.current - n.month, format: :only_month)} :
                                  #{model.monthly_worked_hours(Date.current - n.month)}"
                    end
  if csv
    hours_per_month.reverse.join("\n")
  else
    h.safe_join(hours_per_month.map { |month_total| h.content_tag(:p, month_total) }.reverse)
  end
end
```

This is just an example as a suggestion. I suppose it should work but I don't know. If you do this refactor, write some tests to validate the behaviour

app/admin/members.rb
```diff
@@ -18,16 +18,18 @@
...
     selectable_column
     column :first_name
     column :last_name
     column :role
-    column('3 heures faites?') { |member| member.worked_three_hours?(Date.current) }
     column(:group) do |member|
       group_links = member.groups.map { |group| auto_link group }
       safe_join group_links, ', '
     end
+    column :worked_hours_of_three_last_months
```
Mate2xo: The I18n translation for this column is missing in the BO. You could either add it as an activerecord virtual attribute in the appropriate locale file, or use an admin view locale lookup (such as `column(t '.worked_hours)`)

app/admin/members.rb
```diff
@@ -40,10 +42,8 @@
     column :last_name
     column :phone_number
     column :role
-    column("3 heures de #{l 1.month.ago, format: '%B'}?") { |member| member.worked_three_hours?(1.month.ago) }
-    column("3 heures de #{l Date.current, format: '%B'}?") { |member| member.worked_three_hours?(Date.current) }
-    column("3 heures de #{l 1.month.from_now, format: '%B'}?") { |member| member.worked_three_hours?(1.month.from_now) }
     column(:group) { |member| member.groups.map(&:name).join(', ') }
+    column :worked_hours_of_three_last_months_csv_version
```
Mate2xo: Check that the column has a translated name in the generated CSV.

spec/decorators/member_decorator_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe MemberDecorator do
+end
```
Mate2xo: You could test here your decorators methods, in the case you decide to refactor the two similar methods

spec/features/track_members_worked_hours_spec.rb
```diff
@@ -76,4 +74,13 @@
...
   end
+
+  def set_enrollments_for_hours_resume_test
+    slot = 1.week.ago.clamp(Date.current.at_beginning_of_month, Date.current)
+    3.times do
+      create :enrollment, member: member,
+                          mission: create(:mission, start_date: slot, due_date: slot + 3.hours)
+      slot -= 1.month
+    end
+  end
```
Mate2xo: `resume` in french means either `CV` or `reprendre`. This method name is very confusing and I think it doesn't convey its purpose. Maybe you should rename it as what it does exactly, like `create_enrollments_for_the_last_three_months` (if I understood correctly)
Since that to know what this helper does, we have to scroll down, the name either have to be very explicit or the helper should not exist in the first place (and thus this setup has to be put in a `before` block, since this setup is only used for one given `context`)

spec/models/member_spec.rb
```diff
@@ -180,5 +180,16 @@
         expect(member.monthly_worked_hours(Date.current)).to eq family_member.monthly_worked_hours(Date.current)
       end
     end
+
+    context 'when a member is enrolled on an event' do
+      let(:member) { create :member }
+
+      it "The event's duration are not counted in worked hours" do
```
Mate2xo: "#monthly_worked_hours when a member is enrolled on an event `ignores these hours`"
The convention is to use a verb right after `it`. You can try `rspec --format documentation` to get a feel of how your tests read as a full sentence

Mate2xo (changes requested):

spec/models/member_spec.rb
Floury-Fabio: done

spec/features/track_members_worked_hours_spec.rb
Floury-Fabio: done

spec/decorators/member_decorator_spec.rb
Floury-Fabio: done

app/admin/members.rb
Floury-Fabio: done

app/admin/members.rb
Floury-Fabio: done

app/decorators/member_decorator.rb
Floury-Fabio: done

app/decorators/member_decorator.rb
Floury-Fabio: done

app/admin/members.rb
```diff
@@ -29,7 +29,7 @@
       group_links = member.groups.map { |group| auto_link group }
       safe_join group_links, ', '
     end
-    column :worked_hours_of_three_last_months
+    column(t('.worked_hours')) { |member| MemberDecorator.new(member).hours_worked_in_the_three_last_months }
```
Mate2xo: You shouldn't need to use `MemberDecorator.new(member)` here, as you have already declared the use of the `MemberDecorator` in this ActiveAdmin resource (line 21)
Does it work with just `{ |member| member.hours_worked_in_the_last_three_months }` ?

app/admin/members.rb
```diff
@@ -43,7 +43,7 @@
     column :phone_number
     column :role
     column(:group) { |member| member.groups.map(&:name).join(', ') }
-    column :worked_hours_of_three_last_months_csv_version
+    column(t('active_admin.resource.index.worked_hours')) { |member| MemberDecorator.new(member).hours_worked_in_the_three_last_months(csv: true) }
```
Mate2xo: I think it should also work with just `t '.worked_hours'`. Does it?

app/decorators/member_decorator.rb
```diff
@@ -1,21 +1,24 @@
+# frozen_string_literal: true
+
+# MemberDecorator
 class MemberDecorator < Draper::Decorator
   delegate_all
 
-  def worked_hours_of_three_last_months
+  def hours_worked_in_the_three_last_months(csv: false)
```
Mate2xo: naming -> `hours_worked_in_the_last_three_months`
(sinon ça se lit "heures travaillées sur les 3 mois derniers" (ou "sur les 3 derniers mois de..." selon le contexte) au lieu de "heures travaillées sur les 3 derniers mois")

spec/decorators/member_decorator_spec.rb
```diff
@@ -1,4 +1,30 @@
+# frozen_string_literal: true
+
 require 'rails_helper'
+require 'draper/test/rspec_integration'
 
 RSpec.describe MemberDecorator do
+  describe '#hours_worked_in_the_three_last_months' do
+    let(:generate_member_decorator) { described_class.new(build_stubbed(:member)) }
+
+    context 'when active admin view is generated' do
```
Mate2xo: I think `'when no options are given'` would be a more accurate context, as the MemberDecorator is not aware of the existence of ActiveAdmin. You could for example use it in a regular view.

spec/decorators/member_decorator_spec.rb
```diff
@@ -1,4 +1,30 @@
...
+      it 'return the html version' do
+        member_decorator = generate_member_decorator
+
+        content = member_decorator.hours_worked_in_the_three_last_months
+
+        expect(content).to include('<p>')
+      end
+    end
+
+    context 'when the csv is generated' do
```
Mate2xo: Same as above, `'when the :csv option is true'`

Mate2xo (changes requested):

spec/decorators/member_decorator_spec.rb
Floury-Fabio: done

spec/decorators/member_decorator_spec.rb
Floury-Fabio: done

app/admin/members.rb
Floury-Fabio: don't work with csv

app/admin/members.rb
Floury-Fabio: done

app/decorators/member_decorator.rb
Floury-Fabio: it is already fixed, no ?

Mate2xo (approved):

PR #238, opened 2020-11-09
desactivate confirmation for email update

Floury-Fabio:


Floury-Fabio:
The title tells all.

Mate2xo (changes requested):
This will deactivate email change confirmations for ALL users.
We only want to remove this mail confirmation for 'inactive' users.

With this change, a regular user that wants to change his/her own mail address from his/her profile won't receive any mail confirmation. Thus if the user makes a mistake, he/she won't be able to login anymore

Mate2xo (changes requested):
Just add a controller/request test to frame this particular case and we'll be good to go

Floury-Fabio:
https://github.com/Mate2xo/Creeons_la_coop/issues/231

spec/controllers/members_controller_spec.rb
```diff
@@ -58,8 +58,20 @@
       end
     end
 
-    it "sets a confirmation message" do
-      expect(flash[:notice]).to eq(I18n.translate("activerecord.notices.messages.update_success"))
+    context 'when updating the email' do
+      before { put :update, params: { id: member.id, member: { email: 'update@update.com' } } }
+
+      it "don't update the email directly" do
```
Mate2xo: `dont't` -> `doesn't`

app/admin/members.rb
```diff
@@ -72,10 +72,9 @@ def new_unloggable_member.email_required?
       respond_with_dual_blocks(new_unloggable_member, options, &block)
     end
 
-    def update(&block)
+    def update
       resource.skip_reconfirmation!
-      super do |success, failure|
-        block&.call(success, failure)
+      super do |failure|
         failure.html { render :edit }
```
Mate2xo: You have to use either `|format|` or `|success, failure|` (see [inherited resources](https://github.com/activeadmin/inherited_resources#overwriting-actions))
Here, `failure.html { render :edit }` will actually be rendered whether the update is a failure or not

Mate2xo (changes requested):

Mate2xo (approved):

PR #239, opened 2020-11-10
group_model (bug fix)

Floury-Fabio:
cleaning files organization


Floury-Fabio:
This branch adds a new group management system
The groups have now their own model
Now we can link several group to one member
Each group can have several managers

Mate2xo (approved):

PR #241, opened 2020-12-01
New management of enrollments

Floury-Fabio:
https://github.com/Mate2xo/Creeons_la_coop/issues/226


Mate2xo:
CI Fail

app/admin/missions.rb
```diff
@@ -46,23 +46,38 @@
...
   end
 
   action_item :create_enrollment, only: :show do
-    link_to t('active_admin.new_model', model: Enrollment.model_name.human),
-            new_admin_mission_enrollment_path(resource)
+    if resource.event
+      link_to t('active_admin.new_model', model: Participation.model_name.human),
+        new_admin_mission_participation_path(resource)
+    end
   end
```
Mate2xo, line 81: if the mission is an `event`
  then yes, this action should allow to create a `Participation` associated to this mission and a member
else if the mission is a working session
  then this action should allow to assign an available slot to a member.

This is because some members don't/can't sign up themselves for a given mission; in that case, an admin does the enrollment for that person through this interface

app/admin/participations.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+ActiveAdmin.register Participation do
+  # See permitted parameters documentation:
+  # https://github.com/activeadmin/activeadmin/blob/master/docs/2-resource-customization.md#setting-up-strong-parameters
```
Mate2xo: You can delete this comment. This info is easily accessible on their website anyway

app/admin/slots.rb
```diff
@@ -1,21 +1,27 @@
 # frozen_string_literal: true
 
-ActiveAdmin.register Enrollment do
+ActiveAdmin.register Mission::Slot, as: 'slots' do
```
Mate2xo: Would it cause any problem for the CRUD if we renamed the `Mission::Slots, as: 'slots'` here to `Mission::Slot, as: 'Slot'`?
If possible, I would like to keep the singular resource convention

app/admin/slots.rb
```diff
@@ -1,21 +1,27 @@
...
   # See permitted parameters documentation:
   # https://github.com/activeadmin/activeadmin/blob/master/docs/2-resource-customization.md#setting-up-strong-parameters
```
Mate2xo: You can delete that

app/assets/javascripts/participations.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
```
Mate2xo: Delete this unused file

app/assets/javascripts/slots.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
```
Mate2xo: Delete this unused file

app/assets/stylesheets/participations.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the Participations controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
```
Mate2xo: Delete this unused file

app/assets/stylesheets/slots.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the slots controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
```
Mate2xo: Delete this unused file

app/controllers/missions_controller.rb
```diff
@@ -22,14 +22,22 @@ def create
     generate(@mission)
   end
 
-  def show; end
+  def show
+    @mission = @mission.decorate
+    @mission
```
Mate2xo: I think you can replace these two lines by simply 
```ruby
@mission.decorate
```
and have the same result

But here you can actually remove that code, and use Draper's `decorates_assigned` helper (see [here](https://github.com/drapergem/draper#when-to-decorate-objects))

app/controllers/participations_controller.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+# Manages Members participations on Events
+class ParticipationsController < ApplicationController
+  before_action :authenticate_member!
+  before_action :set_event, only: %i[create destroy]
```
Mate2xo: Apparently, the @event is not used in the `#create` action. You can then remove it from this callback

app/controllers/slots_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# A slot is a place of a mission who can be take by a member. The slots go on 90 minutes.
+# A mission have a limited number of slots( n = duration_of_mission / 90 minutes * members_count).
```
Mate2xo: A Slot is a one fraction of a Mission, that can be taken by a Member. One Slot lasts for 90 minutes.
A Mission has a limited number of slots (n = duration_of_mission / 90minutes * members_count)

app/services/slot/manager.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+# It Generate slot after a mission creation
```
Mate2xo: What's the difference with `Slot::Generator` then? This comment should be much more specific to be helpful, especially as `Manager` is a very generic name that conveys no documentation information in this case. Maybe `Slot::Assigner` or `Slot::Updater` could be ok here?

Floury-Fabio:
Solved

app/decorators/mission_decorator.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+# MissionDecorator
```
Mate2xo: Here, you can do
```ruby
class MissionDecorator < Draper::Decorator # rubocop:disable Style/Documentation
```
instead of writing a useless comment

app/decorators/mission_decorator.rb
```diff
@@ -0,0 +1,26 @@
...
+class MissionDecorator < Draper::Decorator
+  delegate_all
+
+  def search_slots_selectable_by_current_member
+    start_times = object.slots.group(:start_time).count.keys
```
Mate2xo: `.count` is supposed to be an Integer, not a Hash, so how could the `.keys` message work?
Did I miss something, or ActiveRecords allows us to chain after `.count`?

app/decorators/mission_decorator.rb
```diff
@@ -0,0 +1,26 @@
...
+    selectable_slots = []
+
+    start_times.each do |start_time|
+      slot = mission.slots.where(start_time: start_time, member_id: h.current_member.id).first
+      slot = mission.slots.where(start_time: start_time).first  if slot.nil?
```
Mate2xo: is `mission` here actually `object` ?

Mate2xo (changes requested):
First review for now, I'll continue later

app/admin/missions.rb
Floury-Fabio: it's already possible. The subtlety's slot model is the slot count on a mission depends of his max_member and his duration. The slot count change only if this characteristic change(it ' managed by a service object). The slot model is not a only a relation model, it control the enrollment on a mission. A slot represents a place in a mission and places' mission are limited. So we can't generate a row in slot table freely. For create an enrollment, we update a slot which already exist.

app/admin/participations.rb
Floury-Fabio: done

app/admin/slots.rb
Floury-Fabio: done

app/assets/javascripts/participations.coffee
Floury-Fabio: done

app/assets/javascripts/slots.coffee
Floury-Fabio: done

app/assets/stylesheets/participations.scss
Floury-Fabio: done

app/assets/stylesheets/slots.scss
Floury-Fabio: done

app/admin/slots.rb
Floury-Fabio: spelling error
Done

app/controllers/missions_controller.rb
Floury-Fabio: Done.

app/controllers/participations_controller.rb
Floury-Fabio: done

app/controllers/slots_controller.rb
Floury-Fabio: done

app/services/slot/manager.rb
Floury-Fabio: done

app/decorators/mission_decorator.rb
Floury-Fabio: done

app/decorators/mission_decorator.rb
Floury-Fabio: see https://guides.rubyonrails.org/active_record_querying.html#total-of-grouped-items

app/decorators/mission_decorator.rb
Floury-Fabio: is a useless method that forgot to remove

app/decorators/slot_decorator.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# MissionDecorator
+
+class SlotDecorator < ApplicationDecorator
+  delegate_all
+
+  def formated_start_time
```
Mate2xo:
```ruby
def formatted_start_time
```

app/models/mission/slot.rb
```diff
@@ -0,0 +1,18 @@
...
+#
+#  id                :bigint(8)        not null, primary key
+#  start_date        :datetime
+#  mission_id        :bigint()
+#  member_id         :bigint()
+#
+
+# A slot is a place of a mission who can be take by a member. The slots go on 90 minutes.
+# A mission have a limited number of slots( n = duration_of_mission / 90 minutes * members_count).
+class Mission::Slot < ApplicationRecord
```
Mate2xo: A Slot is a one fraction of a Mission, that can be taken by a Member. One Slot lasts for 90 minutes.
A Mission has a limited number of slots (n = duration_of_mission / 90minutes * members_count)

app/services/mission/updator.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def transform_to_event
+    all_is_ok = true
+    Mission::Slot.transaction do
+      @mission.slots.destroy_all
+      @mission.update!(@params)
+    rescue ActiveRecord::RecordInvalid
+      all_is_ok = false
+    end
+    all_is_ok
```
Mate2xo: `.update` is already supposed to return `true` or `false` if the update succeeds/fails, according to validation.
So in theory you could just do
```ruby
    Mission::Slot.transaction do
      @mission.slots.destroy_all
      @mission.update(@params)
    end 
```
(without `!`) to have the same result, and without having to manage any Exception.

Did you do this to make sure the transaction would fail if `.update => false` ?

Mate2xo (changes requested):

Mate2xo:
I'll need more time to review the rest

app/decorators/mission_decorator.rb
Floury-Fabio: solved

app/models/mission/slot.rb
Floury-Fabio: done

spec/requests/missions_spec.rb
```diff
@@ -1,30 +1,21 @@
...
-  let(:create_event) { create :mission, generate_event_attributes }
-  let(:create_mission) { create :mission, generate_mission_attributes }
-  let(:create_mission_with_slots) { create :mission, generate_mission_attributes.merge({ with_slots: true }) }
+RSpec.describe 'A mission request', type: :request do
+  let(:member) { create :member }
 
-  before { sign_in super_admin }
+  before { sign_in member }
 
   describe 'list of all missions' do
```
Mate2xo: So this should actually read 
```ruby
describe 'GET index' do`
```

So now RSpec's `--format documentation` would read `'Mission request GET index has a successful response'`

If you want to test that a list of all Missions is rendered, you should set an explicit expectation in another example for that (e.g. `expect(view.missions.count).to eq 2`

spec/requests/missions_spec.rb
```diff
@@ -1,30 +1,21 @@
...
-                             max_member_count: 4, event: false
-  end
+RSpec.configure do |c|
+  c.include Helpers::SlotEnrollment
+end
 
-  let(:create_event) { create :mission, generate_event_attributes }
-  let(:create_mission) { create :mission, generate_mission_attributes }
-  let(:create_mission_with_slots) { create :mission, generate_mission_attributes.merge({ with_slots: true }) }
+RSpec.describe 'A mission request', type: :request do
```
Mate2xo: `'Mission request'`

spec/requests/missions_spec.rb
```diff
@@ -1,30 +1,21 @@
...
+  let(:member) { create :member }
 
-  before { sign_in super_admin }
+  before { sign_in member }
 
   describe 'list of all missions' do
     it 'successfully renders the list' do
```
Mate2xo: `'has a successful response'`

spec/requests/missions_spec.rb
```diff
@@ -34,29 +25,29 @@
...
+        event = create :mission, event: true
+        create :participation, event_id: event.id, participant_id: member.id
 
         get mission_path(event.id)
 
-        expect(response.body).to include("#{super_admin.first_name} #{super_admin.last_name}")
+        expect(response.body).to include("#{member.first_name} #{member.last_name}")
       end
     end
   end
```
Mate2xo:
```ruby
describe 'GET show' do
  subject(:get_show) { get mission_path(mission) }

  context 'when the mission is an :event' do
    let(:mission) { create_event }
    it 'has a successful response' do
      get_show
      expect(response).to be_successful
    end
    it 'displays the name of the event' do
      ...
    end
    context 'when a member participates in an event' do
      it 'displays the blabla...' do
        create :participation, event_id: mission.id, participant_id: member.id

        get_show

        expect(.....
      end
    end
  end

  context 'when the mission is not an event' do
    it 'returns a successful response' do
      get_show
      expect(response).to be_successful
    end

    ...etc...

  end
end
```

The point here is to regroup REST requests by type, and then you can detail that request with a context. It also allows us to use the `subject` for our `describe`, that is common to all our tests cases. This improves readability, as we have less to read, but the test's story is still obvious with this small abstraction

Mate2xo (changes requested):

Mate2xo:
Again, I'll continue the review on my following break

app/decorators/slot_decorator.rb
Floury-Fabio: done

app/services/slot/assigner.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def assign
+    all_is_ok = true
+    Mission::Slot.transaction do
+      update_slots_in_order_to_remove_these_from_member
+      update_slots_in_order_to_give_these_to_member
+      raise ActiveRecord::Rollback if @errors.any?
+    rescue ActiveRecord::Rollback
+      all_is_ok = false
+    end
```
Mate2xo: If you raise this special error, and define what to do when this error is risen, then I don't think you will benefit from Rails' rollback mechanism. To say it in other words, you are probably overwriting Rails' `ActiveRecord::Rolback` behaviour.
Double check if the transaction is really properly rolled back. If you want to set the `all_is_ok` state, you could put it in a `if` block
e.g.
```ruby
if @errors.any?
  raise ....
  all_is_ok = ...
end
```

app/validators/duration_validator.rb
```diff
@@ -0,0 +1,40 @@
...
+    duration_in_minutes = (mission.due_date - mission.start_date) / 60
+    return true if (duration_in_minutes % 90).zero?
+
+    mission.errors.add :duration, I18n.t('activerecord.errors.models.mission.attributes.duration.multiple')
+    false
+  end
+
+  def duration_extend(mission)
+    duration_in_minutes = (mission.due_date - mission.start_date) / 60
+    return true if duration_in_minutes < 360
```
Mate2xo, line 28: Is there a reason why a mission cannot last more than 6hours? I don't remember seeing it in the specs.
If there is no particular reason, then rather set it to <8h (I think it's the legal limit in France), and explicitly say in the error message that the duration cannot be `greater than 8 hours` instead of `the duration is too high`

app/views/missions/_participants.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <h2 class="heading heading-3 strong-500">
+    <%= t "main_app.views.missions.show.attendees" %> : <%= @mission.participants.length %>
+    <hr class="mt-0 mb-0">
+  </h2>
+
+  <%= render partial: 'participation', collection: @mission.participations %>
+
+  <div class="d-flex justify-content-around">
+    <p>
+    <%= t('.') %> :
```
Mate2xo, line 11: Why is this I18n lookup key empty?

app/views/missions/_participation_form.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= form_with scope: 'participation', url: mission_participations_path(@mission.id) do |f| %>
+  <h4 class="d-block text-center mx-auto">
+    <%= translate ".quick_participation" %>
+  </h4>
+  <%= f.hidden_field :member_id, value: current_member.id %>
+  <% if @mission.participants.select { |m| m.id == current_member.id }.present? %>
```
Mate2xo: If I understood the operations on this line correctly, you can replace it with
```ruby
<% if @mission.participants.include? current_member %>
```

app/views/missions/index.json.jbuilder
```diff
@@ -2,16 +2,20 @@
 
 # JSON formatted for use in the FullCalendar plugin (JS)
 json.array! @missions do |mission|
+  if mission.event
+    members = mission.participants
+  else
+    members = mission.members.uniq
+  end
```
Mate2xo:
```ruby
members = mission.event ? mission.participants : mission.members.uniq
```

Mate2xo (changes requested):

lib/tasks/migrate_for_new_enroll_management.rake
```diff
@@ -0,0 +1,71 @@
...
+  check_enroll_data_consistency
+end
+
+
+def check_enroll_data_consistency
+  puts '****************************************************************************************************'
+  puts '*                               start verification                                                 *'
+  puts '****************************************************************************************************'
+  valid = check_event_data_consistency
+  check_mission_data_consistency if valid
```
Mate2xo: if I understood correctly, you want to lauch `check_mission_data_consistency` if `check_event_data_consistency` has succeeded?

In that case you can write it as
```ruby
check_event_data_consistency && check_mission_data_consistency
```

lib/tasks/migrate_for_new_enroll_management.rake
```diff
@@ -0,0 +1,71 @@
...
+end
+
+def check_event_data_consistency
+
+  valid = true
+  event_enrollments = Enrollment.joins(:mission).where(missions: { event: true })
+  valid = false if event_enrollments.count != Participation.all.count
+  event_enrollments.each do |event_enrollment|
+    result = Participation.find_by(participant_id: event_enrollment.member_id, event_id: event_enrollment.mission_id)
+    valid false if result.nil?
```
Mate2xo: There will be a syntax error here. I thing you meant 
```ruby
valid = false if result.nil?
```

This will do for this PR, as this code should be used only once, but one way to improve this code readability would be to extract the error message rendering to another function (eg. I18n style), so that you could get rid of the `valid` variable, and instead `return` a generated error message when we see data inconsistency

spec/features/mission_events_color_code_spec.rb
```diff
@@ -1,9 +1,23 @@
...
+require 'support/helpers/slot_enrollment.rb'
 
-RSpec.describe "Mission events color codes:", type: :feature do
-  before { sign_in create(:member) }
+RSpec.configure do |c|
+  c.include Helpers::SlotEnrollment
+end
+
+RSpec.describe 'Mission events color codes:', type: :feature do
+  let(:current_member) { create :member, first_name: 'jack' }
```
Mate2xo: This memoized helper is actually not used anywhere (except from the global `before` callback).
It's better here to not use it, and just leave the original `before { sign_in create :member }` to avoid confusion.
Also on line 70, you are creating another member called 'Jack', which seems like a mistake; as on that example, the current member would be Jack, and another member (?) called Jack is created to be enrolled in a mission. This could make make understanding the test's intent difficult if new changes are required later

spec/models/enrollment_spec.rb
```diff
@@ -14,36 +14,4 @@
...
-    end
-
-    it 'returns 0 if any start_time attribute is undefined' do
-      expect(Enrollment.new(end_time: Time.current).duration).to eq 0
-    end
-
-    it 'returns 0 if any end_time attribute is undefined' do
-      expect(Enrollment.new(start_time: Time.current).duration).to eq 0
-    end
-  end
```
Mate2xo, line 48: Are these tests deleted because the `Enrollment` model has been deleted? If so, please delete this file

spec/models/member_spec.rb
```diff
@@ -172,23 +173,30 @@
...
-      let(:family_member) { create :member, register_id: 1234 }
+      let(:create_member) { create :member, register_id: 1234 }
+      let(:create_family_member) { create :member, register_id: 1234 }
+      let(:create_mission_with_slots) { create :mission, name: 'my_mission', max_member_count: 4, event: false, with_slots: true }
 
       it "affects every family members' worked hours count" do
-        create :enrollment, member: family_member
+        member = create_member
+        mission = create_mission_with_slots
+        family_member = create_family_member
```
Mate2xo: These memoized helpers actually do not help us: they just move the readability of the code somewhere else. There are also not used anywhere else than that one and only following test.
Thus they should not exist, and the setup should remain in the test

spec/models/member_spec.rb
```diff
@@ -172,23 +173,30 @@
...
 
-    context 'when a member is enrolled on an event' do
-      let(:member) { create :member }
+    context 'when a member participate on an event' do
+      let(:create_member) { create :member }
 
       it 'ignores these hours' do
-        create :enrollment, member: member
-        create :enrollment, member: member, on_event: true
+        participant = create_member
```
Mate2xo: Same here, this memoized helper is very very useless :smile:

spec/requests/enrollments_spec.rb
```diff
@@ -3,72 +3,4 @@
...
-
-    let(:current_member) { create :member }
-    let(:params) { { enrollment: build(:enrollment, mission: mission, member: current_member).attributes } }
-
-    it "deletes the current members' enrollment" do
-      post mission_enrollments_path(mission), params: params
-
-      disenroll
-    end
-  end
```
Mate2xo, line 73: Do you delete everything here because the `Enrollment` controller does not exist anymore? If so, delete this file please

spec/requests/missions_spec.rb
Mate2xo: You are not testing 'the list' in this test : you are expecting a successful response.
Please change the title of this example as it could become confusing later on

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,405 @@
...
+
+      it 'displays the full name of members who are enrolled in mission' do
+        mission.slots.first.update(member_id: member.id)
+
+        get_show
+
+        expect(response.body).to include("#{member.first_name} #{member.last_name}")
+      end
+
+      context 'when the time slots are already taked' do # rubocop:disable Layout/NestedGroups
```
Mate2xo: -> `'... are already taken'`

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,405 @@
...
+
+  describe 'GET new' do
+    it 'renders successfully the template' do
+      get new_mission_path
+
+      expect(response).to be_successful
+    end
+  end
+
+  describe 'Post' do
```
Mate2xo: `describe 'POST create' do`

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,405 @@
...
+                                              max_member_count: 4, event: false,
+                                              recurrent: true,
+                                              recurrence_rule: IceCube::Rule.daily.to_json,
+                                              recurrence_end_date: DateTime.now + 3.days) }
+        end
+
+        it 'creates the slots for all missions' do
+          lambda_post_mission = -> { post_mission }
+
+          expect { lambda_post_mission.call }.to change { Mission::Slot.count }.by(48)
```
Mate2xo: You don't need a `lambda` here.
You can write
```ruby
           expect { post_mission }.to change { Mission::Slot.count }.by(48) 
```
The bracket notation for `expect` checks for runtime behaviour, and the parenthesis notation ckecks for state

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,405 @@
...
+        end
+      end
+
+      context 'when the duration is extended' do # rubocop:disable Layout/NestedGroups
+        let(:params) { { mission: { name: 'updated_event', due_date: (mission.due_date + 90.minutes) } } }
+
+        it 'adds slots in order to cover the new time slot' do
+          put_mission = -> { update_mission }
+
+          expect { put_mission.call }.to change { mission.reload.slots.count }.by(4)
```
Mate2xo: unnessary `lambda` (same as above)

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,405 @@
...
+        end
+      end
+
+      context 'when the duration is shortened' do # rubocop:disable Layout/NestedGroups
+        let(:params) { { mission: { name: 'updated_event', due_date: (mission.due_date - 90.minutes) } } }
+
+        it 'removes useless slots' do
+          put_mission = -> { update_mission }
+
+          expect { put_mission.call }.to change { mission.reload.slots.count }.by(-4)
```
Mate2xo: unnessary `lambda` (same as above)

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,405 @@
...
+        end
+      end
+
+      context 'when the max_member_count is increased' do # rubocop:disable Layout/NestedGroups
+        let(:params) { { mission: { name: 'updated_event', max_member_count: 5 } } }
+
+        it 'adds slots in order to cover the new count' do
+          put_mission = -> { update_mission }
+
+          expect { put_mission.call }.to change { mission.reload.slots.count }.by(2)
```
Mate2xo: unnessary `lambda` (same as above)

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,405 @@
...
+        end
+      end
+
+      context 'when the max_member_count is reduced' do # rubocop:disable Layout/NestedGroups
+        let(:params) { { mission: { name: 'updated_event', max_member_count: 3 } } }
+
+        it 'removes useless slots' do
+          put_mission = -> { update_mission }
+
+          expect { put_mission.call }.to change { mission.reload.slots.count }.by(-2)
```
Mate2xo: unnessary `lambda` (same as above)

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,405 @@
...
+        end
+
+        it "doesn't remove when slots are occupied" do
+          generate_enrollments_on_n_time_slots_on_a_mission(mission, 4)
+
+          put_mission = -> { update_mission }
+
+          expect { put_mission.call }.not_to(change { mission.reload.slots.count })
```
Mate2xo: unnessary `lambda` (same as above)

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,405 @@
...
+        participants = create_list :member, 4
+        participants.each { |participant| create :participation, event_id: mission.id, participant_id: participant.id }
+
+        delete_mission
+
+        expect(Participation.where(event_id: mission.id)).to be_empty
+      end
+    end
+
+    context 'when the is not an :event' do
```
Mate2xo: 'when the mission is not an event'

Mate2xo (changes requested):

spec/requests/slots_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end
+
+    context 'when the one time slots are unavailable' do
+      let(:slot_params) do
+        { slot: { member_id: member.id,
+                  start_times: [mission.start_date, mission.start_date + 90.minutes] } }
+      end
+
+      it "don't update any slots" do
```
Mate2xo: `'when one of the time slots is unavailable'`
`it 'does not update any slot'`
(if I understood the test correctly)

Mate2xo (changes requested):

Mate2xo (approved):

PR #244, opened 2020-12-10
Static planning

Floury-Fabio:
Add a static slot model and a service who give the possibility to generate automatically a planing for the next month.


app/services/schedule_generator.rb
```diff
@@ -0,0 +1,84 @@
...
+      current_day += 1.day
+    end
+  end
+
+  def generate_missions_for_a_day(current_day)
+    return if current_day.strftime('%a') == 'Sun'
+
+    generate_current_hours(current_day).each do |current_hour|
+      mission = FactoryBot.create :mission, name: 'permanence clac', description: 'permanence clac',
+                                            start_date: current_hour, min_member_count: 2
```
Mate2xo: Do not use FactoryBot for production code. Rather be explicit about what attributes are being set. Otherwise, fiddling with our test suite could have unwanted impact on production.
We only use FactoryBot for tests and seeds.

app/services/schedule_generator.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    members = static_slot.members
+    format_mission_slots(members.count, mission)
+    members.each do |member|
+      slot = mission.slots.find_by(start_time: current_hour)
+      slot.update(member_id: member.id)
+    end
+  end
+
+  def format_mission_slots(members_count, mission)
```
Mate2xo: Could you put a comment here, or find a better name for this method to explain what it does?
I only understand that it updates Missions (I guess that means creating slots according to the `members_count`), I don't know were the 'formatting' role of this method should come from

app/services/schedule_generator.rb
```diff
@@ -0,0 +1,84 @@
...
+    return if mission.max_member_count >= members_count
+
+    params[:max_member_count] = members_count
+    mission_updator = Mission::Updator.new(mission, params)
+    @errors << 'format mission failed' unless mission_updator.call
+  end
+
+  def determine_week_type(current_hour)
+    week_types = %w[D A B C]
```
Mate2xo: Is there a particular reason for this array to be ordered as `D,A,B,C`? Did you need to order it this way, or does the ordering not matter at all?

config/locales/models/static_slot.en.yml
```diff
@@ -0,0 +1,21 @@
+en:
+  activerecord:
+    models:
+      static_slot:
+        one: Static slot
+        other: Static slots
+    attributes:
+      static_slot:
+        hour: heure
```
Mate2xo: `hour`

spec/decorators/static_slot_decorator_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe StaticSlotDecorator do
+end
```
Mate2xo: Add a few quick tests for this decorator please

spec/factories/static_slot_members.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :static_slot_member do
+  end
```
Mate2xo: Is this empty factory valid? I thought that it belonged to `member` and `static_slot`?

spec/models/static_slot_member_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe StaticSlotMember, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
Mate2xo: You can add shouldamatchers `belong_to` quick tests here

spec/models/static_slot_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe StaticSlot, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
Mate2xo: Test validation & associations here please

spec/requests/admin/static_slots_spec.rb
```diff
@@ -0,0 +1,116 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'A StaticSlot request', type: :request do
+  before { sign_in create :member, :super_admin }
+
+  describe 'GET index' do
+    it 'has a successful response' do
+      get admin_static_slots_path
```
Mate2xo, line 12: An empty index page test is near useless, as it should be a standard ActiveAdmin behaviour. You can create a few `StaticSlot` records before the `GET` request to assert that the records do not raise unexpected behaviour

app/models/static_slot.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  enum week_day: { monday: 0, tuesday: 1, wednesday: 2, thursday: 3, friday: 4, saturday: 5, sunday: 6 }
+  enum week_type: { A: 0, B: 1, C: 2, D: 3 }
+
+  validates :week_day, presence: true
+  validates :hour, presence: true, inclusion: 0..23
+  validates :minute, presence: true, inclusion: 0..59
+  validates :week_type, presence: true
+
+  accepts_nested_attributes_for :members
```
Mate2xo: Is this really necessary? I don't think we should allow to modification of `Member`s attributes through a `StaticSlot` form, as we use the `static_slot_members` join table to materialize the association between both.
Do you need to update other attributes to allow a member to register on a static slot?

Mate2xo (changes requested):

app/services/schedule_generator.rb
Floury-Fabio: yes, 
we must have the index 0 for D and index 1 for A because (multiple of 4 modulo 4 == 0)

Mate2xo: Then could you add this implementation detail explanation as a comment please? As it might be otherwise very hard to guess that.

db/migrate/20201211111406_add_hours_to_static_slots.rb
```diff
@@ -0,0 +1,5 @@
+class AddHoursToStaticSlots < ActiveRecord::Migration[5.2]
+  def change
+    add_column :static_slots, :hours, :datetime
```
Mate2xo: Renaming this column `:start_hour` would be more explicit

Mate2xo (changes requested):

app/admin/members.rb
```diff
@@ -58,10 +59,15 @@
           link_to Arbre::Context.new { (status_tag class: 'important', label: group.name) }, [:admin, group]
         end
       end
+      table_for member.static_slots do
+        column 'static_slots' do |static_slot|
```
Mate2xo: Use an I18n translation key here for this column's title.
E.g. `column StaticSlot.model_name.human` that will fetch the `activerecord.models.static_slot.one` locale key

app/admin/static_slots.rb
```diff
@@ -0,0 +1,68 @@
...
+    def update
+      @static_slot = StaticSlot.find(permitted_params[:id])
+      permitted_params = params.require(:static_slot).permit(:week_day, :week_type, :hour, :minute)
+      permitted_params.merge!({ start_time: DateTime.new(2020, 1, 1, permitted_params[:hour].to_i,
+                                                         permitted_params[:minute].to_i) })
+      if @static_slot.update(permitted_params)
+        redirect_to admin_static_slot_path(@static_slot), notice: 'success'
+      else
+        flash[:error] = @static_slot.errors.full_messages.join(', ')
+        render :edit
```
Mate2xo, line 65: I think these methods can be greatly simplified, using tools given by `inherited_resources` gem.
Let's talk about that when we finish this feature

app/jobs/generate_schedule_job.rb
```diff
@@ -0,0 +1,7 @@
+class GenerateScheduleJob < ApplicationJob
+  queue_as :default
+
+  def perform(*args)
+    schedule_generator = ScheduleGenerator.new(args[0])
```
Mate2xo, line 7: you can remove the `schedule_generator = ` part of this line (unused variable)

spec/models/schedule_generator_service_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe ScheduleGenerator, type: :model do
+  describe '#generate_schedule' do
+    subject(:generate_schedule) { ScheduleGenerator.new(current_member).generate_schedule }
+
+    let(:current_member) { create :member }
```
Mate2xo: I think it would be nicer to have this test variable called `author` instead of `current_member`, as it would state explicitly the role the the member passed to the `ScheduleGenerator`: it is used to set the `author` attribute of the created Missions

spec/models/schedule_generator_service_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    before { allow(DateTime).to receive(:current).and_return DateTime.new(2020, 12, 10) }
+
+    it 'generates missions for the next month' do
+      expect { generate_schedule }.to change(Mission, :count).by(78)
```
Mate2xo, line 14: It might be helpful to have a small explanation on why we expect the created Mission count to be `78`

Mate2xo (changes requested):

app/services/schedule_generator.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  def determine_week_type(current_hour)
+    reference = DateTime.new(2020, 9, 7)
+    week_in_seconds = 60 * 60 * 24 * 7
+    week_count_between_reference_and_current_hour = (current_hour.at_beginning_of_week.to_i - reference.to_i) / week_in_seconds
+
+    week_types = %w[D A B C] # we must have the index 0 for D and index 1 for A because (multiple of 4 modulo 4 == 0)
+    week_types[week_count_between_reference_and_current_hour % 4]
+  end
```
Mate2xo: It seems that these methods are not used in this object anymore?

app/services/static_members_recruiter.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def determine_intervals_with_current_time(static_slot)
+    gaps = {}
+    type_of_current_week = determine_week_type(@current_time)
+    gaps['week'] = ((StaticSlot.week_types[static_slot.week_type] + 1) - (StaticSlot.week_types[type_of_current_week] + 1 )) % 4
+    gaps['day'] = ((StaticSlot.week_days[static_slot.week_day] + 1) - @current_time.strftime('%u').to_i) % 7
+    gaps['week'] -= 1 if (StaticSlot.week_days[static_slot.week_day] + 1) < @current_time.strftime('%u').to_i
+    (gaps['week'].weeks + gaps['day'].days)
+  end
```
Mate2xo, line 63: Could you put an explanatory comment here please? I have no idea what it does and why

db/migrate/20201213213101_create_history_of_generated_schedules.rb
```diff
@@ -0,0 +1,9 @@
+class CreateHistoryOfGeneratedSchedules < ActiveRecord::Migration[5.2]
+  def change
+    create_table :history_of_generated_schedules do |t|
+      t.datetime :month_of_generated_schedule
```
Mate2xo: This might be better shortened to `month_number`?

Mate2xo (changes requested):

Mate2xo (approved):

PR #245, opened 2020-12-14
Static planning: from a member's point of vue

Floury-Fabio:
https://github.com/Mate2xo/Creeons_la_coop/issues/225


app/controllers/members_controller.rb
```diff
@@ -17,9 +18,13 @@ def edit
   end
 
   def update
+    if params['member']['static_slots_attributes'].present?
+      Member::StaticSlotAssigner.call(current_member, permitted_params['static_slots_attributes'])
+    end
+
```
Mate2xo: The User should see a flash message if his/her static slot assignement has failed for whatever reason.

Since there are now two distinct steps in this `#update` action, and you would want to set errors feedback messages according to a given step, you might like to use [this gem](https://dry-rb.org/gems/dry-transaction/0.13/) to manage this through 'Railway programming' style.

So you could create a 'Transaction' object, to call from this action. The result of the transaction would then guide the action.
In the 'Transaction' object, you'll have 2 steps: `assign_static_slots` and `update_attributes`. If one step fails, this step returns a `Failure` with a custom Hash containing whatever info you need (e.g. `return Failure({errors: 'this is a custom error message})`)
The controller action would behave according to that Transaction's result

e.g.:
```ruby
def update
  transaction = Members::UpdateTransaction.call(params: permitted_params)
  if transaction.success?
      flash[:notice] = t 'activerecord.notices.messages.update_success'
      redirect_to member_path(@member.id) 
  else
      flash[:error] = transaction[:errors]
      render :edit 
  end
end
```

Using this, the controller stays simple, and the feedback messages setup logic is moved to the Transaction object.
I would suggest to put this transaction object in `app/services/transactions`

app/services/member/static_slot_assigner.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# It updates slots before a mission update
+class Member::StaticSlotAssigner < ApplicationService # rubocop:disable Style/ClassAndModuleChildren
```
Mate2xo: Follow Rubocop's recommendation here.
You can use this namespace:
```ruby
module Members # plural
  class StaticSlotAssigner < ApplicationService
    ...
  end
end
```

app/services/member/static_slot_assigner.rb
```diff
@@ -0,0 +1,27 @@
...
+  def initialize(current_member, static_slot_params)
+    @current_member = current_member
+    @static_slot_ids = extract_ids(static_slot_params)
+
+    @errors = []
```
Mate2xo: It seems that this `@errors` is not used?

config/locales/views/mission.en.yml
```diff
@@ -18,6 +18,7 @@ en:
       participate_button: Participate in this event
       stop_participate_button: Stop participate in this event
     enrollment_form:
-      detailled_slot: De %{start_time} à %{end_time}
       button_enroll_to_slots: update enrollment
-      unavailability_time_slot: There is no more time slot available.
+      detailled_slot: De %{start_time} à %{end_time}
+      fixed_member: You have a static slot. Please see an admin if you want change your time slot.
+      unavailability_time_slot: There is no more time slot available
```
Mate2xo: please rename the key as: `unavailability_time_slot` -> `no_time_slot_available`

config/locales/views/mission.en.yml
```diff
@@ -18,6 +18,7 @@ en:
       participate_button: Participate in this event
       stop_participate_button: Stop participate in this event
     enrollment_form:
-      detailled_slot: De %{start_time} à %{end_time}
       button_enroll_to_slots: update enrollment
-      unavailability_time_slot: There is no more time slot available.
+      detailled_slot: De %{start_time} à %{end_time}
```
Mate2xo: please rename the key as: `detailled_slot` -> `slot_details`

config/locales/views/missions.fr.yml
```diff
@@ -18,6 +18,7 @@ fr:
       participate_button: Participer à cet événement
       stop_participate_button: Ne plus participer à cet évévenement
     enrollment_form:
-      detailled_slot: De %{start_time} à %{end_time}
       button_enroll_to_slots: Mettre à jour l'inscription
-      unavailability_time_slot: Il n y a plus de créneau disponible pour cette mission.
+      detailled_slot: De %{start_time} à %{end_time}
+      fixed_member: Un créneau fixe vous été attribué. Veuillez voir un admin si vous souhaitez changer de créneau.
+      unavailability_time_slot: Il n y a plus de créneau disponible pour cette mission
```
Mate2xo: Same as for the related `en` I18n locale

Mate2xo (changes requested):

app/transactions/members/update_transaction.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class Members::UpdateTransaction
+  include Dry::Transaction
+
+  step :extract_ids
```
Mate2xo: Since this step will never fail, you can use `tee` instead of `step`, to signify that this operation will happen everytime
```ruby
tee :extract_ids
```

app/transactions/members/update_transaction.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+    Success(input.merge({ static_slot_ids: static_slot_ids }))
+  end
+
+  def assign_static_slot(input)
+    return Success(input) if input[:permitted_params][:static_slots_attributes].nil?
+
+    input[:static_slot_ids].each do |static_slot_id|
+      unless ::StaticSlotMember.create(member_id: input[:current_member].id, static_slot_id: static_slot_id)
+        Failure(error: t('activerecord.errors.models.mission.messages.static_slot_extraction_failure'))
```
Mate2xo: I guess that you meant `static_slot_creation_failure` instead of `static_slot_extraction_failure` ?

By the way, is it possible to use the I18n shorthand `t('.static_slot_creation_failure')` here?

app/transactions/members/update_transaction.rb
```diff
@@ -0,0 +1,41 @@
...
+    Success(input)
+  end
+
+  def update_member(input)
+    input[:permitted_params].delete('static_slots_attributes')
+    unless input[:current_member].update(input[:permitted_params])
+      Failure(t('activerecord.errors.messages.update_fail'))
+    end
+
+    Success(input[:current_member])
```
Mate2xo: `Success(input)` should also be fine here. Is there a reason why you chose to only leave the `:current_member` key from the `input`?

config/locales/active_admin.en.yml
```diff
@@ -65,6 +65,7 @@ en:
         remove_static_slots_of_this_member: 'Remove the static slots from this member'
         week: 'week'
       edit:
+        remove_static_slots_of_this_member: 'Retirer les créneaux fixes de ce membre'
```
Mate2xo: English here please

config/locales/views/mission.en.yml
```diff
@@ -18,6 +18,7 @@ en:
       participate_button: Participate in this event
       stop_participate_button: Stop participate in this event
     enrollment_form:
-      detailled_slot: De %{start_time} à %{end_time}
       button_enroll_to_slots: update enrollment
-      unavailability_time_slot: There is no more time slot available.
+      fixed_member: You have a static slot. Please see an admin if you want change your time slot.
+      no_time_slot_available: There is no more time slot available
+      slot_details: De %{start_time} à %{end_time}
```
Mate2xo: English localization here

Mate2xo (changes requested):

app/transactions/members/update_transaction.rb
Mate2xo: ?

Mate2xo (changes requested):

config/locales/models/static_slot.en.yml
```diff
@@ -19,3 +15,12 @@ en:
           Saturday: Saturday
           Sunday: Sunday
         week_type: week type
+    errors:
+      models:
+        mission:
+          messages:
+            static_slot_attribution_failure: static_slot_attribution_failed
```
Mate2xo: no camel case for user feedback please

config/locales/models/static_slot.fr.yml
```diff
@@ -19,3 +15,12 @@ fr:
           Saturday: Samedi
           Sunday: Dimanche
         week_type: type de semaine
+    errors:
+      models:
+        mission:
+          messages:
+            static_slot_attribution_failure: Echec de l attribution des créneaux fixes
```
Mate2xo: apostrophe

Mate2xo (changes requested):

Mate2xo (approved):

PR #254, opened 2020-12-18
Static member options

Floury-Fabio:
https://github.com/Mate2xo/Creeons_la_coop/issues/247

### DIFF
--- a/app/models/history_of_static_slot_selection.rb
+++ b/app/models/history_of_static_slot_selection.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# This model keeps the selections of static slots by the members.
+# It serves to prevent the abuses by the members.
+class HistoryOfStaticSlotSelection < ApplicationRecord
+  belongs_to :member
+  belongs_to :static_slot
+end
--- a/app/transactions/members/update_transaction.rb
+++ b/app/transactions/members/update_transaction.rb
@@ -6,6 +6,7 @@ class Members::UpdateTransaction
   tee :params
   tee :extract_ids
   step :assign_static_slot
+  step :save_in_history
   step :update_member
 
   private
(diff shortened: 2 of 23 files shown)


app/models/history_of_static_slot_selection.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# This model keeps the selections of static slots by the members.
+# It serves to prevent the abuses by the members.
```
Mate2xo, line 4: `# This history is used to monitor misuses from enrolling members`

app/transactions/members/update_transaction.rb
```diff
@@ -3,39 +3,56 @@
...
   step :update_member
 
   private
 
-  def extract_ids(input)
-    return Success(input) if input[:permitted_params][:static_slots_attributes].nil?
+  def params(input)
+    @current_member = input[:current_member]
+    @static_slots_attributes = input[:permitted_params][:static_slots_attributes]
+    input[:permitted_params].delete('static_slots_attributes')
```
Mate2xo, line 17: You should be able to do
```ruby
    @static_slots_attributes =  input[:permitted_params].delete('static_slots_attributes') 
```
since `.delete` actually returns the deleted value

app/transactions/members/update_transaction.rb
```diff
@@ -3,39 +3,56 @@
...
-    static_slot_ids = []
-    input[:permitted_params][:static_slots_attributes].values.each do |value|
-      static_slot_ids << value['id']
+    @static_slot_ids = []
+    @static_slots_attributes.each_pair do |_key, value|
+      @static_slot_ids << value['id']
     end
-    input.merge!({ static_slot_ids: static_slot_ids })
-    Success(input)
+    @static_slot_ids.uniq
```
Mate2xo, line 28: what is the point of this?
If you want to mutate the `@static_slot_ids` value, you should do `.uniq!`

app/transactions/members/update_transaction.rb
```diff
@@ -3,39 +3,56 @@
...
+    Success
```
Mate2xo, line 29: Pass the input hash here: `Success(input)`, or else the following steps won't have acess to that hash

app/views/layouts/_static_slot_fields.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="card m-2 p-2 nested-fields">
   <div class="col-lg-12">
     <div class="field">
-      <%= f.label :test, 'static_slot' %><br />
+      <%= f.label :static_slot, t('activerecord.models.static_slot.one') %><br />
```
Mate2xo, line 4: Use `StaticSlot.model_name.human` instead of directly querying the i18n key. It's a bit more readable, and maybe easier to maintain

Mate2xo (changes requested):
Since you are closing this PR, I won't finish the review, but I leave the comment if it can be of any use

Floury-Fabio:
The slot model has been revert. This branch require the slot model, so this pull request is closed. 
Restart is here : https://github.com/Mate2xo/Creeons_la_coop/tree/revert_slot_management
and here: https://github.com/Mate2xo/Creeons_la_coop/tree/add_regulated_mission.

PR #255, opened 2020-12-18
restore old label

Floury-Fabio:
https://github.com/Mate2xo/Creeons_la_coop/issues/252


Floury-Fabio:
The slot model has been revert. This branch require the slot model, so this pull request is closed. 
Restart is here : https://github.com/Mate2xo/Creeons_la_coop/tree/revert_slot_management
and here: https://github.com/Mate2xo/Creeons_la_coop/tree/add_regulated_mission.

PR #259, opened 2020-12-29
Revert slot management

Floury-Fabio:
Revert project at 799b8f0(before slot management introduction). The next branch do same features without this model and more easily.
For the review i recommend a git diff with 799b8f0.

next branch: https://github.com/Mate2xo/Creeons_la_coop/tree/add_regulated_mission


Mate2xo:
/!\ CI fail /!\
Is that to be expected?

Mate2xo (approved):

PR #261, opened 2020-12-29
Add regulated mission

Floury-Fabio:
Add regulated mission in order to control enrollments on them


app/controllers/enrollments_controller.rb
```diff
@@ -6,11 +6,14 @@ class EnrollmentsController < ApplicationController
...
   def create
-    return flash[:alert] = translate('.max_member_count_reached') if member_slots_full?
+    create_transaction = Enrollments::CreateTransaction.new.with_step_args(
+      validate: [mission: @mission],
+      prepare: [mission: @mission]
+    ).call(permitted_params)
 
-    @enrollment = Enrollment.create(enrollment_params)
-    user_feedback_on_creation
+    return flash[:alert] = create_transaction.failure unless create_transaction.success?
```
Mate2xo: Are you setting the full `failure` object as the error message?
Or is the message properly set on transaction failure?

app/controllers/missions_controller.rb
```diff
@@ -27,11 +27,15 @@ def show; end
...
+    update_transaction = Missions::UpdateTransaction.new.with_step_args(
+      update: [mission: @mission]
+    ).call(permitted_params)
+
+    if update_transaction.success?
       flash[:notice] = translate 'activerecord.notices.messages.update_success'
       render :show
     else
-      flash[:error] = translate 'activerecord.errors.messages.update_fail'
+      flash[:error] = update_transaction.failure
```
Mate2xo, line 34: Same question as above

app/controllers/missions_controller.rb
```diff
@@ -66,16 +70,36 @@ def generate(mission)
...
+  def permit_params_of_regulated_mission
+    params.require(:mission).permit(
+      :name, :description, :event, :delivery_expected,
+      :recurrent, :recurrence_rule, :recurrence_end_date,
+      :max_member_count, :min_member_count,
+      :due_date, :start_date, :genre,
+      addresses_attributes: %i[id postal_code city street_name_1 street_name_2 _destroy],
+      enrollments_attributes: [:id, :_destroy, :member_id, start_time: []]
+    )
+  end
```
Mate2xo: Since the only difference with the standard params is only in the `enrollments_attributes` hash key, you could do something like that to simplify:

```ruby
def regulated_mission_params
  base_params.merge(enrollments_attributes: [:id, :_destroy, :member_id, start_time: []])
end

def standard_mission_params
  base_params.merge(enrollments_attributes: %i[id _destroy member_id start_time end_time])
end

def base_params
    params.require(:mission).permit(
      :name, :description, :event, :delivery_expected,
      :recurrent, :recurrence_rule, :recurrence_end_date,
      :max_member_count, :min_member_count,
      :due_date, :start_date, :genre, 
end
```
to make it more readable.

Just make sure that the merged params are properly permitted, but this is a kind of pattern that I've used before

Mate2xo (changes requested):

app/models/mission.rb
```diff
@@ -31,14 +32,48 @@ class Mission < ApplicationRecord
...
+  end
+
+  private
+
+  def time_slot_available?(current_time_slot, member)
+    member_enrollment = enrollments.find_by(mission: self, member: member)
+    if member.nil? || member_enrollment.nil?
+      return enrollments.where('start_time <= :current_time_slot AND :current_time_slot < end_time',
+                               current_time_slot: current_time_slot)
+                        .count < max_member_count
```
Mate2xo: indentation

app/decorators/enrollment_decorator.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class EnrollmentDecorator < ApplicationDecorator # rubocop:disable Style/Documentation
+  delegate_all
+
+  def already_taken?(time_slot)
+    return false if enrollment.id.nil?
```
Mate2xo: I think you can use `enrollment.new_record?` here, to be more readable (if that's your intention here)

app/models/mission.rb
```diff
@@ -31,14 +32,48 @@ class Mission < ApplicationRecord
...
+    end
+
+    return true if member.present? && member_enrollment.decorate.already_taken?(current_time_slot)
```
Mate2xo: Something seems wrong here,  I am confused. I think the semantics must be refactored.

The code says "A time_slot is available if it is between a given member's enrollment begin and end" ... But then, if a given member is already enrolled, why is this time slot available for him?
If you want to know if a given time slot is available for anyone to take, I guess you should ask the Model directly, not a given record for a given member

The semantics say "A time_slot is available if a given member exists, and if this member's enrollement is already taken". Sorry I don't understand what that actually means

Either way, I have no idea of what the real intention is.

`def available_time_slots` suggests that you want to query all available time slots for a given mission. the `member` argument suggests that you want to query available time slots for a given member. `member = nil` default value suggest that this method can return all available time slots for a given mission, or all available time slots for a given member, or both (it is not explicit).

Whatever it is you want to do, I feel that you should separate the responsabilities. Because without you to explain to me how this work, I would not be able to refactor this if needed (I would then delete everything and start from scratch)

Let's talk about it when we can

Mate2xo (changes requested):

app/transactions/enrollments/create_transaction.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      failure_message = I18n.t('enrollments.create.max_member_count_reached')
+      Failure(failure_message)
+    end
+
+    def prepare(input, mission:)
+      return Success(input) unless mission.genre == 'regulated'
+
+      input['end_time'] = input['start_time'].last.to_datetime + 90.minutes
+      input['start_time'] = input['start_time'].first.to_datetime
```
Mate2xo: This transaction has knowledge it shouldn't have:
It knows that the end_time should be based on the last `input['start_time']`
It knows that the start_time should be based on the first `input['start_time']`
This knowledge creates an invisible dependency, that is invisible to the caller : the caller has no way to know that a specific array order is needed in order to properly do the transaction.

You should use dependency injection here, as giving an explicit argument to the transaction. In doing so, you won't need the `prepare` step anymore.

You would then create an Enrollement by choosing explicit items from the `input` hash

Mate2xo (changes requested):

app/controllers/enrollments_controller.rb
Floury-Fabio: The message is rendered properly on view

app/controllers/missions_controller.rb
Floury-Fabio: The message is rendered properly on view

app/transactions/enrollments/create_transaction.rb
Floury-Fabio: If i inject an explicit argument, i prepare this argument in controller. The aim of transaction is reduce controller. 
If i sort start_time in transaction, is acceptable?

Mate2xo: Yeah that's a good intent. But the Controller 'controls' what's going on. Other objects are responsible for the actions they are asked to do, and depending and what they are asked to do, we may pass them different arguments. So it's better to give an object what it needs for a given action, rather that giving that object the knowledge of WHO is asking the action.
E.g.: here, the transaction knows that the asked action comes from a given action in a given controller. So this context could be different if that action was asked in another action from an another controller.

I think it's a good idea to see the interaction as a discussion between the controller and the transaction:
- Controller: Hey could you make me an enrollement record?
- Transaction: Yeah, on what `mission` ?
- Controller: On `this` one please
- Transaction: And on what time sould it start, and what time should it end?
- Controller: From 10am to 1pm
- Transaction: OK done

Sorting the given array would be better yes, but  the problem stays the same: you are asking the Transaction to search for a given record ; but you also asked it to create an enrollment.
So It's better if you give the Transaction what it needs to work explicitely. As a nice side effect, that would document the use of this transaction: someone reading the code in the controller would know what this transaction needs (e.g.: e given record, an array, or a hash...), without knowing how it works.

That can allow us, for example, to have the minimum amount of information needed to modify the controller as we want, without modifying the transaction's behaviour without knowing it (eg.: by changing some data structures that the transaction would not recognize)

Mate2xo: I guess it's actually the same logic as using `.with_step_args` : you explicitely pass the argument you want the transaction to work with

app/models/mission.rb
```diff
@@ -17,11 +17,17 @@
...
 #
 
 # A Mission is an activity that has to be done for the Supermaket Team to function properly.
 # Every member can create a mission
 # Available methods: #addresses, #author, #due_date, #name, #description
+# A regulated mission have several time_slots
+# A time slot is a subdivision of mission duration
+# A time slot last 90 minutes
+# A time slot have several slots
+# Slots count for a time slot is equal to :max_member_count
```
Mate2xo, line 31: Great comment, thanks

app/transactions/missions/update_transaction.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      Success(input)
+    end
+
+    def prepare_params(input)
+      return Success(input) if input[:genre] != 'regulated'
+
+      input['enrollments_attributes'].each do |_key, enrollment|
+        enrollment['end_time'] = enrollment['start_time'].last.to_datetime + 90.minutes
+        enrollment['start_time'] = enrollment['start_time'].first
```
Mate2xo: Same as with `enrollments/create_transaction.rb`

app/transactions/missions/update_transaction.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      Success(input)
+    end
+
+    def update(input, mission:)
+      if mission.update(input)
+        Success(input)
+      else
+        failure_message = "#{I18n.t('activerecord.errors.messages.update_fail')}
+        #{mission.errors.full_messages.join(', ')}"
```
Mate2xo: Indent the second line of the string please.
Check if the message is rendered properly on the user side, because the indentation you use in the code could have an impact on the rendering
e.g.: 
```ruby
        "#{I18n.t('activerecord.errors.messages.update_fail')}
        #{mission.errors.full_messages.join(', ')}"
```
might give 
```
First part of the message
        second part of the message
```

You might need to use a Ruby Heredoc (by using `<<-` or `<<~`, see [this guide](https://www.rubyguides.com/2018/11/ruby-heredoc/))

app/views/missions/_form.html.erb
```diff
@@ -95,17 +100,11 @@
...
-            <%= link_to_add_association t(".enroll_another_member"),
-                                        f,
-                                        :enrollments %>
-          </div>
-        </div>
+        <% if mission.genre == "regulated"%>
+          <%= render 'enrollments_with_time_slots_form', f: f, mission: mission%>
+        <% else %>
+          <%= render 'enrollments_form', f: f, mission: mission %>
+        <% end %>
```
Mate2xo: The only thing that differs between `_enrollments_with_time_slots_form.html.erb` and `_enrollments_form.html.erb` partials, it the sub-partial that they call.
You can then simplify the partials logic by merging `_enrollments_with_time_slots_form.html.erb` and `_enrollments_form.html.erb` together, and pass the name of the sub-partial you want to render in this form.
E.g.:

```ruby
<% if mission.genre == 'regulated' %>
  <%= render 'enrollments_form',
             f: f,
             mission: mission,
             cocoon_fields_name: 'enrollments_with_time_slots_field' %>
<% else %>
  <%= render 'enrollments_form',
             f: f,
             mission: mission,
             cocoon_fields_name: 'enrollments_fields' %>
<% end %>
```
and then using the `coccon_fields_name` variable in the enrollments_form accordingly, to render the desired 'enrollments fields' partial, and target the desired 'enrollments fields' partial for the add/remove_association link

config/locales/views/mission.en.yml
```diff
@@ -1,9 +1,20 @@
...
+      slot_details: From %{start_time} to %{end_time}
+    form:
+      enrolled_members: Enrolled members
+      genre: Genre
+    regulated_quick_enrollment_form:
+      quick_enroll: Quick enroll
+    standard_quick_enrollment_form:
       quick_enroll: Quick enroll
+    update:
+      time_slot_requirement: At least one time slot must be indicated
```
Mate2xo: `At least one time slot must be given`

lib/tasks/change_time_type_of_enrollments.rake
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+desc 'change time type of enrollments attributes'
+task change_time_type_of_enrollments: :environment do
+  puts '****************************************************************************************************'
+  puts '*                               start migration                                                    *'
+  puts '****************************************************************************************************'
```
Mate2xo: If you want, you can use Ruby's `String#ljust`, `String#rjust`, and `String#center`.
e.g.:
```ruby
puts ''.ljust 80, '*'
puts 'start migration'.center 80
puts ''.ljust 80, '*'
```
if you find it easier to write

spec/models/mission_spec.rb
```diff
@@ -52,4 +61,36 @@
       it { is_expected.to have_and_belong_to_many(:addresses) }
     end
   end
+
+  describe '#selectable_time_slots' do
+    it 'returns the time slots with at least one slot available' do
+      mission = create :mission, genre: 'regulated'
+
+      response = mission.selectable_time_slots
```
Mate2xo: do not use `response` as the variable name, as it can shadow the variable `response` set by RSpec in some situations (e.g. in request tests, when you make a `get` request, RSpec creates a `response` variable that you can access to inspect what was the response of the request)

It shouldn't cause any problem in a unit test (for this example here, there is no problem), but I think it's better to avoid to keep a good habit.

FYI: other variables names you should avoid in RSpec for the same reasons: `request`, `subject` (there might be other, but I don't remember them right now)

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+
+  before { sign_in create :member }
+
+  describe 'Get show' do
+    subject(:get_mission) { get mission_path(mission) }
+
+    context 'when the genre is set to standard' do
+      let(:mission) { create :mission }
+
+      it 'rends the standard partial' do
```
Mate2xo: `it 'renders`[...]

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+
+      mission_params.each do |key, value|
+        expect(Mission.last.attributes[key.to_s]).to eq value
+      end
+    end
+
+    context 'when the genre is set to event' do
+      let(:mission_params) { attributes_for(:mission, genre: 'event') }
+
+      it 'creates the mission with the related params' do
```
Mate2xo: `it 'creates the mission with the specified params' do `

spec/requests/missions_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+      end
+
+      it 'adds member enrollment' do
+        put mission_path(mission.id), params: { mission: mission_params }
+
+        enrollment_expected_params.each do |key, value|
+          expect(mission.enrollments.first[key]).to eq value
+        end
+      end
+    end
```
Mate2xo:
```ruby
context 'when enrollment params are given ' do
  # put here examples for a 'normal' mission (=not regulated)
  ...

  context 'with a :regulated mission do
    # put here examples for a :regulated mission
  end
end
```

I think writing it like this would frame the 'enrollment' update better. What do you think?

Mate2xo (changes requested):

app/models/mission.rb
Floury-Fabio: indentation is ok, no ?

app/transactions/enrollments/create_transaction.rb
Floury-Fabio: done

app/transactions/missions/update_transaction.rb
Floury-Fabio: done

spec/models/mission_spec.rb
Floury-Fabio: ok ty

spec/requests/missions_spec.rb
Floury-Fabio: i agree with you

Mate2xo (approved):

Mate2xo (approved):

PR #265, opened 2021-01-06
add a task which creates some static slots

Floury-Fabio:
Add a task which creates some static slots in order to facilitate the work of admin.

### DIFF
--- a/lib/tasks/create_static_slots.rake
+++ b/lib/tasks/create_static_slots.rake
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+desc 'create some static slots'
+task create_some_static_slots: :environment do
+  common_days = %w[Wednesday Thursday Friday]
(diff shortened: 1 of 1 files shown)


lib/tasks/create_static_slots.rake
```diff
@@ -0,0 +1,43 @@
...
+    create_static_slots_for_one_saturday(week_type)
+  end
+  puts 'creation done'
+end
+
+def create_static_slots_for_one_day(week_type, week_day)
+  create_static_slots_for_one_morning(week_type, week_day)
+  create_static_slots_for_one_afternoon(week_type, week_day)
+end
+
```
Mate2xo, line 22: Leave a comment to tell future readers that the date is actually not important here, and that this task is only a one-time use

Mate2xo (changes requested):

Mate2xo (approved):

PR #268, opened 2021-01-07
Recurrent change option for mission

Floury-Fabio:
Add a recurrent change option for missions.

If this option is enable, the admin update will change all mission with the same hours, same week day and the same type planned after the updated mission.

https://github.com/Mate2xo/Creeons_la_coop/issues/248

### DIFF
--- a/app/admin/missions.rb
+++ b/app/admin/missions.rb
@@ -88,7 +88,7 @@ ActiveAdmin.register Mission do
       end
     end
 
-    # helpers
+    private
 
     def generate_update_transaction
       Admin::Missions::UpdateTransaction
--- a/app/admin/missions.rb
+++ b/app/admin/missions.rb
@@ -11,7 +11,8 @@ ActiveAdmin.register Mission do
                 :min_member_count,
                 :start_date,
                 :due_date,
-                :cash_register_proficiency_requirement
+                :cash_register_proficiency_requirement,
+                :recurrent_change
 
   index do
     selectable_column
@@ -40,6 +41,7 @@ ActiveAdmin.register Mission do
       f.input :cash_register_proficiency_requirement,
               :as => :select,
               collection => Mission.cash_register_proficiency_requirements
+      f.input :recurrent_change, as: :boolean if f.object.persisted?
     end
 
     actions
@@ -63,8 +65,8 @@ ActiveAdmin.register Mission do
     panel 'Participants' do
       table_for resource.enrollments do
         column :member
-        column(:start_time) do |enrollment| enrollment.start_time.strftime('%H:%M') end
-        column(:end_time) do |enrollment| enrollment.end_time.strftime('%H:%M') end
+        column(:start_time) { |enrollment| enrollment.start_time.strftime('%H:%M') }
+        column(:end_time) { |enrollment| enrollment.end_time.strftime('%H:%M') }
         column 'actions' do |enrollment|
           link_to(t('active_admin.edit'), edit_admin_mission_enrollment_path(mission, enrollment)) +
             ' ' +
--- a/app/admin/missions.rb
+++ b/app/admin/missions.rb
@@ -76,6 +76,28 @@ ActiveAdmin.register Mission do
     end
   end
 
+  controller do
+    def update
+      if update_transaction.success?
+        redirect_to admin_mission_path(resource.id)
+      else
+        render :edit
+      end
+    end
+
+    # helpers
+
+    def update_transaction
+      Admin::Missions::UpdateTransaction
+        .new
+        .with_step_args(
+          update_mission: [mission: resource],
+          get_updatable_missions: [old_mission: resource]
+        )
+        .call({ params: permitted_params[:mission] })
+    end
+  end
+
   action_item :create_enrollment, only: :show do
     link_to t('active_admin.new_model', model: Enrollment.model_name.human),
             new_admin_mission_enrollment_path(resource)
--- a/app/admin/missions.rb
+++ b/app/admin/missions.rb
@@ -78,7 +78,6 @@ ActiveAdmin.register Mission do
 
   controller do
     def update
-      update_transaction = generate_update_transaction
       if update_transaction.success?
         flash[:notice] = translate 'activerecord.notices.messages.update_success'
         redirect_to admin_mission_path(resource.id)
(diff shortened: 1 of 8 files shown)


app/admin/missions.rb
```diff
@@ -74,6 +76,31 @@
...
+      if update_transaction.success?
+        flash[:notice] = translate 'activerecord.notices.messages.update_success'
+        redirect_to admin_mission_path(resource.id)
+      else
+        flash[:error] = update_transaction.failure
+        render :edit
+      end
+    end
+
+    # helpers
```
Mate2xo: use `private` here

app/controllers/missions_controller.rb
```diff
@@ -27,6 +27,7 @@ def show; end
   def edit; end
 
   def update
+    update_transaction = generate_update_transaction
```
Mate2xo: try memoization

app/transactions/admin/missions/update_transaction.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      around :transaction
+
+      step :update_mission
+      tee :get_updatable_missions
+      step :update_recurrently_other_missions
+
+      private
+
+      def transaction(input, &block)
```
Mate2xo: Rename this method to reflect intent. E.g. `rollback_if_failure`

Mate2xo (changes requested):

app/admin/missions.rb
```diff
@@ -74,6 +76,33 @@
...
+        flash[:error] = update_transaction.failure
+        render :edit
+      end
+    end
+
+    private
+
+    def update_transaction
+      @update_transaction ||=
+        begin
```
Mate2xo, line 94: Why do you need a `begin` here?

app/controllers/missions_controller.rb
```diff
@@ -72,10 +72,13 @@ def generate(mission)
   end
 
   def update_transaction
-    Missions::UpdateTransaction.new.with_step_args(
-      transform_time_slots_in_time_params_for_enrollment: [regulated: @mission.regulated?],
-      update: [mission: @mission]
-    ).call(permitted_params)
+    @update_transaction ||=
+      begin
```
Mate2xo, line 76: same here, why `begin`?

app/transactions/admin/missions/update_transaction.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      # helpers
+
+      def updatable_missions(old_mission)
+        Mission.all.select do |mission|
+          mission.start_date > old_mission.start_date &&
+            mission.start_date.strftime('%R%u') == old_mission.start_date.strftime('%R%u') &&
+            mission.genre == old_mission.genre
+        end
```
Mate2xo: `Mission` is the biggest resource in this app. Do NOT put all existing records in memory all at once: ask the DB to give you only the records you want with SQL before processing data with Ruby.
Also, I think maybe the only mission that may be interesting to change in this case would be the `regulated` Missions. Did Caroline asked that also `events` must be recurrently updated?

app/transactions/admin/missions/update_transaction.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      def update_recurrently_other_missions(input)
+        return Success(input) unless input[:params][:recurrent_change]
+
+        failure_message = I18n.t('activerecord.errors.messages.update_fail')
+
+        input[:missions].each do |mission|
+          return Failure(failure_message) unless mission.update(input[:params])
+        end
```
Mate2xo, line 62: If I understand correctly:
* lets say there are 3 missions on Monday 7, Monday 14, Monday 21 (of January)
* we update Monday 7 to have a new title

=> Since we are using Monday7's `params`, Monday14 and Monday21 will also see their date change to January 7th. I'm pretty sure that's not what we want

The associated ticket is very vague and needs to be much more specific. Let's take some time to re-spec this feature

spec/requests/admin/missions_spec.rb
```diff
@@ -23,10 +23,49 @@
...
+
   def create_history_of_generated_schedule_for_n_months(months_count)
     (1..months_count).each do |n|
       create :history_of_generated_schedule,
              month_number: (DateTime.current + n.month).at_beginning_of_month
     end
   end
+
+  def create_same_missions_planned_after_this_mission(mission)
+    occurrente_date = mission.start_date + 7.days
```
Mate2xo: `occurence_date`

spec/requests/admin/missions_spec.rb
```diff
@@ -23,10 +23,49 @@
...
+  describe 'PUT' do
+    subject(:put_mission) { put admin_mission_path(updated_mission.id), params: { mission: mission_params } }
+
+    before { allow(DateTime).to receive(:current).and_return DateTime.new(2020, 2, 3, 9) }
+
+    context 'when the recurrent changes params is true' do
+      let(:updated_mission) { create :mission, start_date: DateTime.current + 2.days }
+      let(:mission_params) { { name: 'updated_mission', recurrent_change: true } }
+
+      it 'changes the missions with the same week day, the same hour, same genre and who are planned after the updated mission' do # rubocop:disable Layout/LineLength
```
Mate2xo: `it 'updates future missions that match the same week day, hour, and genre'`

spec/requests/admin/missions_spec.rb
```diff
@@ -23,10 +23,49 @@
...
+  end
+
   def create_history_of_generated_schedule_for_n_months(months_count)
     (1..months_count).each do |n|
       create :history_of_generated_schedule,
              month_number: (DateTime.current + n.month).at_beginning_of_month
     end
   end
+
+  def create_same_missions_planned_after_this_mission(mission)
```
Mate2xo: `create_future_matching_missions`

Mate2xo (changes requested):

app/admin/missions.rb
Floury-Fabio: see https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/
in multi-line memoization section

app/controllers/missions_controller.rb
Floury-Fabio: see https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/
in multi-line memoization section

spec/requests/admin/missions_spec.rb
Mate2xo: Still not corrected (english fix)

spec/requests/admin/missions_spec.rb
Mate2xo: still not fixed (english)

Mate2xo (changes requested):

Mate2xo (approved):

PR #271, opened 2021-01-08
patch recruiter

Floury-Fabio:
Fix bug with merge enroll in recruiter object.

This method merge now only the enrollment with the same member_id and the same mission_id


Floury-Fabio:
branch is renamed in new branch

PR #276, opened 2021-01-14
Fix bug on event in admin views

Floury-Fabio:
We need admins to be able to edit the `:genre` attributes of Mission records.
Since the role of the `:event` attribute has now been taken by the `:genre` enum, we also need to convert existing records to use this `:genre` attribute.

* add Mission `:genre` management in admin views
* the `conversion:mission_event_bool_to_genre_enum` task will convert existing mission records' `:event` boolean to `:genre` enum

### DIFF
--- a/lib/tasks/convert_event_bool_in_enum_genre.rake
+++ b/lib/tasks/convert_event_bool_in_enum_genre.rake
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+# This namespace contain all tasks who adapt the data at changements of db structure
+namespace :convertion do # rubocop:disable Metrics/BlockLength
+  desc 'it set enum genre at event for mission who have the event boolean at true'
+  task convert_event_bool_in_enum_genre: :environment do
+    puts ''.ljust 80, '*'
+    puts 'start migration'.center 80
+    puts ''.ljust 80, '*'
+    file = File.open(Rails.root.join('db/convert_event_bool_in_genre_enum/convert_event_bool_in_genre_enum.sql'))
+    ActiveRecord::Base.connection.execute(file.read)
+    file.close
+    puts ''.ljust 80, '*'
+    puts 'migration done'.center 80
+    puts ''.ljust 80, '*'
+
+    puts ''.ljust 80, '*'
+    if check_data_consistency_for_event_bool_convertion
+      puts 'migration success'.center 80
+    else
+      puts 'migration failure'.center 80
+    end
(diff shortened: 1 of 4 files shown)


lib/tasks/convert_event_bool_in_enum_genre.rake
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# This namespace contain all tasks who adapt the data at changements of db structure
+namespace :convertion do # rubocop:disable Metrics/BlockLength
```
Mate2xo: `conversion`

lib/tasks/convert_event_bool_in_enum_genre.rake
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# This namespace contain all tasks who adapt the data at changements of db structure
```
Mate2xo: This namespace contain`s` all tasks  `that` adapt `existing data when the DB structure changes`

lib/tasks/convert_event_bool_in_enum_genre.rake
```diff
@@ -0,0 +1,36 @@
...
+namespace :convertion do # rubocop:disable Metrics/BlockLength
+  desc 'it set enum genre at event for mission who have the event boolean at true'
```
Mate2xo: it set`s` the :genre enum to the value of 2 (= event) for missions that have the :event boolean attribute set to true

lib/tasks/convert_event_bool_in_enum_genre.rake
```diff
@@ -0,0 +1,36 @@
...
+  task convert_event_bool_in_enum_genre: :environment do
```
Mate2xo: task mission_event_bool_to_genre_enum

spec/requests/admin/missions_spec.rb
```diff
@@ -23,6 +25,20 @@
     end
   end
 
+  describe 'POST' do
+    subject(:post_mission) { post admin_missions_path, params: { mission: mission_params } }
+
+    context 'when the :genre params is set to event' do
+      let(:mission_params) { attributes_for :mission, genre: 'event', author_id: current_admin.id }
+
+      it 'creates the mission with the genre setted to event' do
```
Mate2xo: the genre `set` to event

Mate2xo (changes requested):

Mate2xo (approved):

PR #277, opened 2021-01-15
Blog restoration

Floury-Fabio:
We need a news feed. This branch restore an old feature without the interface elements which already ensured by active admin.
The news feed is managed by the info model.

- move index which was in infos controllers documents in documents controllers
- remove useless methods from infos controllers
- restore link to infos index view in navbar
- add request infos tests for index and show


Floury-Fabio:
Some features is missing.

PR #279, opened 2021-01-21
Blog restoration

Floury-Fabio:
https://github.com/Mate2xo/Creeons_la_coop/issues/223


app/models/group.rb
```diff
@@ -16,4 +16,11 @@ class Group < ApplicationRecord
   has_many :members, through: :group_members
 
   validates :name, presence: true, uniqueness: true
+
+  def self.localized_name_of_group_which_grant_rights_to_members(name_to_localize)
+    list_of_groups_with_rights = ['redactor']
+    return nil unless list_of_groups_with_rights.include?(name_to_localize)
+
+    I18n.t("activerecord.attributes.group.localized_names.#{name_to_localize}")
+  end
```
Mate2xo: If I understand correctly, in the end this method is intended to be used only with the argument `'redactor'`.
Why not use `Group.human_enum_name('redactor')` then?

app/models/info.rb
```diff
@@ -6,13 +6,20 @@
...
 #  title      :string
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
 #  author_id  :bigint(8)
 #
 
 class Info < ApplicationRecord
-  validates :title, presence: true
+  extend Enumerize
+  extend ActiveModel::Naming
```
Mate2xo: Why do you need this ?
It should be available by default, as `Info` already is an `ActiveModel`. Is it because you saw it in the Enumerize documentation? If that's the case, this option is needed only for a plain Ruby object, not a Rails object

app/models/member.rb
```diff
@@ -86,6 +86,13 @@ def monthly_worked_hours(date)
       .reduce(0.0) { |sum, enrollment| sum + enrollment.duration }
   end
 
+  def redactor?
+    groups.each do |group|
+      return true if group.name == Group.localized_name_of_group_which_grant_rights_to_members('redactor')
```
Mate2xo: Same as above, why not use `#human_enum_name`?

db/schema.rb
```diff
@@ -239,6 +241,16 @@
...
+  create_table "temp_enrollments_from_csv", id: false, force: :cascade do |t|
+    t.bigint "member_id"
+    t.bigint "mission_id"
+    t.bigint "id"
+    t.time "old_start_time"
+    t.time "old_end_time"
+    t.datetime "start_time"
+    t.datetime "end_time"
+  end
+
```
Mate2xo: I guess this table comes from another ticket, does it?

lib/tasks/remove_old_infos.rake
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# This namespace contains all tasks related to the infojmodel
+namespace :info do # rubocop:disable Metrics/BlockLength
+  desc 'removes the infos whose the last update dateback to more than one month'
```
Mate2xo, line 5: `'deletes Info records that are more than one month old'`

spec/factories/members.rb
```diff
@@ -47,5 +51,12 @@
 
     trait :admin do role { 'admin' } end
     trait :super_admin do role { 'super_admin' } end
+
+    after :create do |member, options|
+      if options.redactor?
+        group = create :group, name: Group.localized_name_of_group_which_grant_rights_to_members('redactor')
```
Mate2xo: Same as above, maybe `#human_enum_name`?

spec/features/document_uploads_at_info_section_spec.rb
```diff
@@ -26,17 +26,17 @@
   context "when an admin uploads a document" do
```
Mate2xo, line 26: I think we could rename this file as `documents_uploads_at_document_page_spec.rb`, since the `Info` and `Document` now have their own dedicated pages

Mate2xo (changes requested):

app/models/info.rb
Floury-Fabio: It is a stain of one try that i forgot to remove.
I remove them.

db/schema.rb
Floury-Fabio: yes it come from https://github.com/Mate2xo/Creeons_la_coop/tree/restore_enrollments_with_csv
it is removed.

Mate2xo (approved):

PR #287, opened 2021-02-15
Feedback from jobs

Floury-Fabio:
Add action cable channel in order to generate notifications after the enroll static members job.

For that we do:

- add redis for production
- add i18n js gem
- config notifications channel
- rewrite recruiter service for collect reports
- config active admin for call of cable and i18n_js


app/services/static_members_recruiter.rb
```diff
@@ -106,4 +118,27 @@ def interval_in_day_between_current_time_and_next_day_of_same_type(static_slot)
     rank_of_current_day = @current_time.strftime('%u').to_i
     (rank_of_static_slot_day - rank_of_current_day) % 7
   end
+
+  def find_mission_with_adequate_cash_register_requirement(member, missions, time_slot)
```
Mate2xo: `def find_adequate_mission_or_send_error_report(...)`

app/transactions/members/update_transaction.rb
```diff
@@ -9,7 +9,7 @@ class UpdateTransaction
     tee :extract_static_slots_ids
     step :assign_static_slot
     step :save_in_history
-    step :enroll_member_according_to_new_static_slots
+    tee :enroll_member_according_to_new_static_slots
```
Mate2xo: `    tee :enroll_member_according_to_new_static_slots_or_send_error_report`

app/assets/javascripts/application.js
```diff
@@ -55,3 +55,4 @@
 //= require thredded-custom
 //= require flash_messages_on_xhr
 //= require js.cookie
+//= require i18n/translations
```
Mate2xo: Is this really necessary as we use js translation only in the admin interface? Delete it otherwise

app/assets/javascripts/channels/notifications.js
```diff
@@ -0,0 +1,18 @@
+App.notifications = App.cable.subscriptions.create("NotificationsChannel", {
+  received(data) {
+    let html = `<div class='flash flash_alert'>
+                  <p> ${I18n.t('admin.members.js.end_of_auto_enrollment')} </p>
```
Mate2xo: If this is the only place that you use I18n in JS, you could maybe try to rename this file into `notifications.js.erb` and see if the asset pipeline compiles it properly.
If it does, then you could do
```
                  <p> <%= I18n.t('admin.members.js.end_of_auto_enrollment') %> </p>
```
and get rid of the `i18n-js` gem entirely

spec/models/static_members_recruiter_service_spec.rb
```diff
@@ -48,4 +48,56 @@
       expect(mission.members).to include(member)
     end
   end
+
+  context 'when all static slots of all members has a corresponding time slot in missions' do
```
Mate2xo: `  context 'when all static slots of all members have been assigned to a corresponding time slot in the upcoming missions' do`

spec/models/static_members_recruiter_service_spec.rb
```diff
@@ -48,4 +48,56 @@
...
+      create_members_with_static_slot
+      create_missions
+
+      recruiter = recruit_static_members
+
+      expect(recruiter.reports.size).to eq 0
+    end
+  end
+
+  context 'when a static slots of one member has not a corresponding time slot in missions' do
```
Mate2xo: ` context 'when a static slot of a given member has no available time slot in upcoming missions' do`

spec/models/static_members_recruiter_service_spec.rb
```diff
@@ -48,4 +48,56 @@
...
+    end
+
+    let(:create_members_with_static_slot) do
+      static_slot = create :static_slot, week_day: 'Monday', week_type: 'A', start_time: DateTime.new(2021, 1, 4, 9)
+      create :member_static_slot, member: member, static_slot: static_slot
+      static_slot2 = create :static_slot, week_day: 'Tuesday', week_type: 'A', start_time: DateTime.new(2021, 1, 5, 9)
+      create :member_static_slot, member: member, static_slot: static_slot2
+    end
+
+    it 'The recruiter has no reports' do
```
Mate2xo: Copy/paste fail :1st_place_medal:

Mate2xo (changes requested):

spec/models/static_members_recruiter_service_spec.rb
Floury-Fabio: :)

app/admin/members.rb
```diff
@@ -126,7 +126,9 @@
   end
 
   collection_action :enroll_static_members, method: :post do
-    EnrollStaticMembersJob.perform_later
+    # We must wait the end of page reload for keep job notifications
```
Mate2xo: `    # We must wait the end of page reload from the redirection. Otherwise if the job finishes before the page reload, the user won't get the 'job finished' notification from ActionCable`

Mate2xo (approved):

PR #290, opened 2021-03-22
Add validation checks for participations and missions requests

Floury-Fabio:
This branch add validation checks for admin missions requests and admin enrollment requests

For mission, it checks:
 - if related enrollments are not outside the new mission's period
 - if related enrollments on regulated mission match the mission's time slot

For enrollment, it checks:
- if the standard mission is full
- the slot availability for regulated mission
- the cash register proficiency for the regulated mission
- the positivity of the duration
- if the enrollment's date times are inside the period of the mission

We add also a drop down preset for creating enrollments, and fix display errors when rendering the response of a mission `#create` request.


config/locales/models/enrollment.en.yml
```diff
@@ -0,0 +1,12 @@
+en:
+  activerecord:
+    errors:
+      models:
+        enrollment:
+          inconsistent_datetimes: The selected datetimes are outside of the period of the mission
```
Floury-Fabio, line 11: alphabetical order

app/admin/enrollments.rb
```diff
@@ -9,13 +9,71 @@
...
+      @update_enrollment_transaction ||=
+        begin
+          mission = Mission.find(params[:mission_id])
+          member = Member.find(permitted_params[:enrollment][:member_id])
+          input = permitted_params[:enrollment].merge({ mission: mission, member: member, id: params[:id] })
+          Admin::Enrollments::UpdateTransaction
+            .new
+            .call(input)
+        end
+    end
```
Mate2xo: Since you use `build_resource`, if the available `resource` helper already contains the associated `mission` and `member`, you could simplify this by doing something like
```ruby
def create
  build_resource
  transaction_result = Admin::Enrollments::CreateTransaction.new.call(resource)
  if transaction_result.success?
    flash[:notice] = translate 'enrollments.create.confirm_enroll'
    redirect_to admin_mission_path(params[:mission_id])
  else
    flash[:error] = transaction_result.failure
    render :new
  end
end
```

app/admin/missions.rb
```diff
@@ -83,9 +83,20 @@
   end
 
   controller do
+    def create # rubocop:disable Metrics/AbcSize
+      build_resource
+      if resource.save
+        flash[:notice] = translate 'missions.activerecord.notices.messages.record_created'
```
Mate2xo: For all admin controller I18n lookups, do:
`translate '.message_that_you_what_to_lookup'`
This will use an I18n shortcut such like `admin.resource_name_pluralized.controller_action_name`

So in this case that would be for example:
`flash[:notice] = t '.success'`
and this would lookup `admin.missions.create.success`
and you could put the locale file in `config/locale/controllers/admin/missions.fr.yml`

app/admin/missions.rb
```diff
@@ -83,9 +83,20 @@
...
+      else
+        flash[:error] = resource.errors.full_messages.join
+        render :new
+      end
+    end
+
     def update
       if update_transaction.success?
-        flash[:notice] = translate 'activerecord.notices.messages.update_success'
+        flash[:notice] = translate 'missions.update.confirm_update'
```
Mate2xo: Same here, try to use  the `t '.your_key'` shortcut where you can to help tidying locales with the default convention

app/transactions/admin/missions/recurrent_update_transaction.rb
```diff
@@ -0,0 +1,83 @@
...
+        Mission.transaction do
+          result = block.call(Success(input))
+          raise ActiveRecord::Rollback if result.failure?
+
+          result
+        end
+        result
+      end
+
+      def remove_datetime_attributes(input)
```
Mate2xo: Leave a comment for why this step is necessary
e.g.:
`# For recurrent changes, we only update non time-related attributes to not mess with individual time slots validations. Time slots validation's responsibility belongs to each individual missions`

app/transactions/admin/missions/recurrent_update_transaction.rb
```diff
@@ -0,0 +1,83 @@
...
+               name: mission.name,
+               start_date: mission.start_date,
+               failure: failure)
+      end
+
+      def build_mission_input(mission, params)
+        mission_params = mission.attributes
+        params.each do |key, value|
+          mission_params[key] = value
+        end
```
Mate2xo: `params[:the_attribute_you_want] = mission.the_attribute_you_want`
Choose explicitly the attributes that you want to setup

Mate2xo (changes requested):

app/transactions/admin/enrollments/update_transaction.rb
```diff
@@ -0,0 +1,180 @@
...
+      end
+
+      def check_cash_register_mastery(input)
+        return Success(input) unless input[:enrollment].mission.regulated?
+
+        failure_message = I18n.t('activerecord.errors.models.enrollment.insufficient_cash_register_mastery')
+        return Failure(failure_message) unless inspect_all_time_slots_for_register_cash_mastery(input)
+
+        Success(input)
+      end
```
Mate2xo: Since this is duplicated from the related `create_transaction`, and this actually checks if an `Enrollment` is valid, you should extract these methods and add them as validation steps into the `Entrollment` model

app/transactions/admin/missions/recurrent_update_transaction.rb
```diff
@@ -0,0 +1,95 @@
...
+               failure: failure)
+      end
+
+      def build_mission_input(mission, params)
+        params[:start_date] = mission.start_date
+        params[:due_date] = mission.due_date
+        { params: params, mission: mission }
+      end
+
+      def remove_datepicker_params(input, key)
```
Mate2xo: `      def remove_activeadmin_datepicker_params(input, key)`

app/transactions/admin/missions/update_transaction.rb
```diff
@@ -6,72 +6,82 @@ module Missions
...
-      def remove_datetime_attributes_if_recurrent_option(input)
-        return Success(input) unless input[:params][:recurrent_change]
-
-        input[:params].delete(:start_date)
-        input[:params].delete(:due_date)
-
+      def check_if_enrollments_match_a_mission_s_time_slots_for_regulated_mission(input)
+        enrollments = input[:mission].enrollments
+        i18n_key = 'activerecord.errors.models.mission.mismatch_between_time_slots_and_related_enrollments'
+        failure_message = I18n.t(i18n_key)
```
Mate2xo:
```ruby
failure_message = I18n.t('mismatch_between_time_slots_and_related_enrollments',
                              scope: %i[activerecord errors models mission])
```

app/transactions/admin/missions/update_transaction.rb
```diff
@@ -6,72 +6,82 @@ module Missions
...
+        enrollments.each do |enrollment|
+          return Failure(failure_message) unless match_a_time_slot?(input, enrollment)
+        end
         Success(input)
       end
```
Mate2xo: THese methods are actually validating the `Mission` object ; so move them as validation steps in the `Mission` model.
By  doing so, we might be able to get rid of this transaction object.

config/locales/controllers/enrollments.fr.yml
```diff
@@ -6,6 +6,9 @@ fr:
       insufficient_proficiency: Votre maîtrise de la caisse est insuffisante pour prendre la dernière place du créneau de %{start_time} à %{end_time}
       max_member_count_reached: Le nombre maximum de participants est déjà atteint
       time_slots_requirement: Ils faut indiquer des créneaux fixes
```
Floury-Fabio: 'il faut'

Mate2xo (changes requested):

app/models/enrollment.rb
```diff
@@ -28,6 +28,90 @@ def contain_this_time_slot?(time_slot)
...
+
+    proficiency_level_of_member = Member.cash_register_proficiencies[member.cash_register_proficiency]
+    unless mission.slot_available_for_given_cash_register_proficiency?(self, proficiency_level_of_member)
+      failure_message = I18n.t('activerecord.errors.models.enrollment.insufficient_cash_register_proficiency')
+      errors.add :insufficient_cash_register_proficiency, failure_message
+      return false
+    end
+
+    true
+  end
```
Mate2xo: These methods should be `private`.
Then, to call them when you want to validate an `Enrollment` instance, use Rails' `validate` helper ; as in
```ruby
validate :check_cash_register_proficiency
``` 
Then, instead of asking your `enrollment` object in the `Admin::Enrollment::CreateTransaction` every `check_...` methods, you could just do `enrollment.valid?` to populate the `.errors`

app/models/mission.rb
```diff
@@ -89,6 +95,76 @@ def available_slots_count_for_a_time_slot(time_slot)
...
+    return nil unless regulated?
+
+    current_time_slot = enrollment.start_time
+    while current_time_slot < enrollment.end_time
+      return false if available_slots_count_for_a_time_slot(current_time_slot).zero?
+
+      current_time_slot += 90.minutes
+    end
+    true
+  end
```
Mate2xo: This might be better in the `Enrollment` model, as you will then ask each associated `enrollement` if they are still valid.
See also the Rails API documentation on the `has_many` options, as there are some validation-related options

app/models/enrollment.rb
```diff
@@ -28,6 +28,90 @@ def contain_this_time_slot?(time_slot)
...
+  def duration_multiple_of_90_minutes?
+    ((end_time.to_i - start_time.to_i) % (60 * 90)).zero?
+  end
+
+  def check_if_datetimes_of_enrollment_are_inside_the_mission_s_period
+    unless mission.inside_period?(self)
+      failure_message = I18n.t('activerecord.errors.models.enrollment.inconsistent_datetimes')
+      errors.add :inconsistent_datetimes, failure_message
+      return false
+    end
```
Mate2xo: Since an Enrollment belongs to a Mission, and you are asking here to the Mission if the current enrollment is valid; the responsiblity should be kept in the Enrollment.
You could refactor this as
```ruby
  def check_if_datetimes_of_enrollment_are_inside_the_mission_s_period
    unless inside_mission_period?
      failure_message = I18n.t('activerecord.errors.models.enrollment.inconsistent_datetimes')
      errors.add :inconsistent_datetimes, failure_message
      return false
    end
  end

  private

  def inside_mission_period?
    start_time >= mission.start_date &&
      start_time <= mission.due_date &&
      end_time >= mission.start_date &&
      end_time <= mission.due_date
  end

**Review by Mate2xo — CHANGES_REQUESTED** [review]

File: app/transactions/admin/enrollments/create_transaction.rb [hunk 15]
```diff
@@ -0,0 +1,114 @@
...
+      end
+
+      def check_cash_register_proficiency(input)
+        return Success(input) unless input.mission.genre == 'regulated'
+
+        failure_message = I18n.t('activerecord.errors.models.enrollment.insufficient_cash_register_mastery')
+        return Failure(failure_message) unless slot_available_for_given_cash_register_proficiency?(input)
+
+        Success(input)
+      end
```
Mate2xo: All these methods have already been moved to the `Enrollement`. You should delete them

app/transactions/admin/enrollments/create_transaction.rb
```diff
@@ -0,0 +1,114 @@
...
+        mission = input.mission
+        member = input.member
+
+        mastery_level_of_member = Member.cash_register_proficiencies[member.cash_register_proficiency]
+
+        proficiency_requirement = mission.cash_register_proficiency_requirement
+        mastery_level_of_mission = Mission.cash_register_proficiency_requirements[proficiency_requirement]
+
+        mastery_level_of_member >= mastery_level_of_mission
+      end
```
Mate2xo: It seems that this method isn't called anywhere?

app/models/mission.rb
```diff
@@ -89,6 +95,76 @@ def available_slots_count_for_a_time_slot(time_slot)
...
+    failure_message = I18n.t('mismatch_between_time_slots_and_related_enrollments',
+                             scope: %i[activerecord errors models mission])
+    enrollments.each do |enrollment|
+      unless match_a_time_slot?(enrollment)
+        errors.add :mismatch_between_time_slots_and_related_enrollments, failure_message
+        return false
+      end
+    end
+    true
+  end
```
Mate2xo: Move them to `private`, and call them through the `validate :your_method` Rails helper.
You could then call `.valid` on your `Mission` instance in `Admin::Missions::UpdateTransaction`

app/transactions/admin/enrollments/update_transaction.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      def prepare_object(input)
+        enrollment = input[:enrollment]
+        params = input[:params]
+
+        params.keys.each do |current_key| # rubocop:disable Style/HashEachMethods
+          enrollment[current_key] = params[current_key] if enrollment.attributes.keys.include?(current_key)
+        end
```
Mate2xo: Are you trying to do the same thing as `.assign_attributes` here ?

app/transactions/admin/enrollments/update_transaction.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+
+      def validation(input) # rubocop:disable Metrics/AbcSize
+        enrollment = input[:enrollment]
+        enrollment.check_if_the_standard_mission_is_not_full
+        enrollment.check_if_the_duration_is_positive
+        enrollment.check_if_datetimes_of_enrollment_are_inside_the_mission_s_period
+        enrollment.check_if_enrollment_is_matching_the_mission_s_timeslots
+        enrollment.check_slots_availability_for_regulated_mission
+        enrollment.check_cash_register_proficiency
```
Mate2xo: Use `.valid?` instead

app/transactions/admin/missions/update_transaction.rb
```diff
@@ -6,72 +6,59 @@ module Missions
...
-        else
-          failure_message = I18n.t('activerecord.errors.messages.update_fail')
-          Failure(failure_message)
+      def prepare_object(input)
+        mission = input[:mission]
+        params = input[:params]
+
+        params.keys.each do |current_key| # rubocop:disable Style/HashEachMethods
+          mission[current_key] = params[current_key] if mission.attributes.keys.include?(current_key)
         end
```
Mate2xo: `mission.assign_attributes(params)`

app/transactions/admin/missions/update_transaction.rb
```diff
@@ -6,72 +6,59 @@ module Missions
...
       end
 
-      def get_updatable_missions(input, old_mission:)
-        return Success(input) unless input[:params][:recurrent_change]
+      def validation(input)
+        mission = input[:mission]
+        mission.check_if_enrollments_are_inside_new_mission_s_period
+        mission.check_if_enrollments_match_a_mission_s_time_slots_for_regulated_mission
+        return Failure(mission.errors.values.flatten[0]) if mission.errors.present?
```
Mate2xo: use `.valid?`

config/locales/models/mission.en.yml
```diff
@@ -16,6 +16,8 @@ en:
     errors:
       models:
         mission:
+          inconsistent_datetimes_for_related_enrollments: At last one enrollment is outside the mission's period
+          mismatch_between_time_slots_and_related_enrollments: Mismatch_between_time_slots_and_related_enrollments
```
Mate2xo: The `recurrent_update_cancel` key is missing (compared to the .fr locale)

spec/requests/admin/enrollments_spec.rb
```diff
@@ -0,0 +1,390 @@
...
+  # this helper imitate the struture of date_params created by active admin datepicker
+  def convert_datetime_in_params(datetime, key)
+    {
+      "#{key}(1i)": datetime.year,
+      "#{key}(2i)": datetime.month,
+      "#{key}(3i)": datetime.day,
+      "#{key}(4i)": datetime.hour,
+      "#{key}(5i)": datetime.min
+    }
+  end
```
Mate2xo: I think I saw this same helper at least 3 times. You could maybe extract it in its own service object

Mate2xo (changes requested):

spec/requests/admin/enrollments_spec.rb
```diff
@@ -0,0 +1,390 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'A Enrollment admin request', type: :request do
```
Mate2xo: Since all the nested `describe` are HTTP method, I think that having the main parent `describe` as 
```ruby
RSpec.describe 'admin/enrollments/', type: :request do
```
would be more explicit. What do you think?

spec/requests/admin/enrollments_spec.rb
```diff
@@ -0,0 +1,390 @@
...
+    end
+
+    it 'confirms the enrollment creation' do
+      post_enrollment
+      follow_redirect!
+
+      expect(CGI.unescapeHTML(response.body)).to include(I18n.t('enrollments.create.confirm_enroll'))
+    end
+
+    context 'when the related mission is :regulated and the enrollment match a time slot' do
```
Mate2xo: context 'when the related mission is :regulated and the enrollment match`es` a time slot' do

spec/requests/admin/enrollments_spec.rb
```diff
@@ -0,0 +1,390 @@
...
+        assign_other_members_to_this_mission
+
+        post_enrollment
+
+        expect(response.body).to include(I18n.t('activerecord.errors.models.enrollment.slot_unavailability'))
+      end
+    end
+
+    context 'when the related mission is regulated, there is one slot left,
+             and the member has an insufficient cash register mastery' do
```
Mate2xo: mastery -> proficiency

spec/requests/admin/missions_spec.rb
```diff
@@ -29,32 +28,241 @@
...
 
+      let(:all_missions) { create_future_matching_missions(mission) + [mission] }
+      let!(:expected_start_dates) { all_missions.map(&:start_date) }
+      let!(:expected_due_dates) { all_missions.map(&:due_date) }
+
       it 'updates futures missions that match the same week day, hour, and genre' do # rubocop:disable Layout/LineLength
-        other_missions = create_future_matching_missions(updated_mission)
+        other_missions = create_future_matching_missions(mission)
 
         put_mission
```
Mate2xo, line 289: no expectation?

Mate2xo (changes requested):

app/admin/missions.rb
```diff
@@ -83,29 +83,33 @@
   end
 
   controller do
+    def create # rubocop:disable Metrics/AbcSize
+      build_resource
+      if resource.save
+        flash[:notice] = translate '.mission_created'
+        redirect_to admin_mission_path(resource.id)
+      else
+        flash[:error] = resource.errors.full_messages.join
```
Floury-Fabio: resource.errors.full_messages.join(', ')

Mate2xo (approved):

Issue #316, opened 2022-05-30
Rename documents


CymanTrachap:
Add a feature to rename a document already uploaded instead of having to delete it to replace it with one renamed locally first

Issue #315, opened 2022-05-30
Move documents from one category/sub-category to another


CymanTrachap:
Add a feature to ease documents re-classification by moving files from one category to another after they've been uploaded

PR #329, opened 2023-01-16
Add update method to document controller

Creiwry:
Tickets: #315 and #316

Added features:
- edit document view
- update method in documents controller
  - > update name of Activestorage::blob
  - > update category
- locales for documents view
- rspec test for update method


app/controllers/documents_controller.rb
```diff
@@ -37,7 +52,17 @@ def destroy
...
+  end
+
+  def update_transaction
+    @update_transaction ||=
+      begin
+        Documents::UpdateTransaction.new.with_step_args(
+          change_filename: [document: @document],
+          update_file: [document: @document]
+        ).call(permitted_params)
+      end
```
Mate2xo: * Why did you need to memoize this ?
* I don't find it very useful to use `step_args` here, as you pass the same argument to each steps, and the value of these arguments are also the same. I would prefer if you just use `.new.call(permitted_params.merge(document: @document))`, as that would make it obvious that the transaction will work with the same document

app/transactions/documents/update_transaction.rb
```diff
@@ -0,0 +1,40 @@
...
+        else
+          failure_message = <<-MESSAGE
+            "#{I18n.t('activerecord.errors.messages.update_fail')}
+            #{document.errors.full_messages.join(', ')}"
+          MESSAGE
+          Failure(failure_message)
+        end
+      else
+        Success(params)
+      end
```
Mate2xo: This method is reasonably long. Try to split it into different private methods, so that the reader can understand what it does at a glance. That might give you some ideas on how to reduce this `if` nesting

config/locales/views/documents.fr.yml
```diff
@@ -0,0 +1,6 @@
+fr:
+  documents:
+    file_name: Nom du document
```
Mate2xo: It doesn't seem that this I8n key is actually used, as you use instead the lookup `'activerecord.attributes.document.filename'`. Also, the `en` and `fr` keys for this value are different.
If that's an attribute translation that you want, you can use `Document.human_attribute_name(:file_name)`

app/views/documents/edit.html.erb
```diff
@@ -0,0 +1,44 @@
...
+              </br>
+                <div class="field">
+                  <%= form.label :category, t('activerecord.attributes.document.category')%><br />
+                  <%= form.select :category, Document.category.options, :selected => @document.category, class: "form-control form-control-lg" %>
+                </div>
+                <br/>
+
+                <div class="d-flex justify-content-between flex-wrap">
+                  <div class="field flex-fill">
+                    <%= form.label :file_name, t('activerecord.attributes.document.filename') %>
```
Mate2xo: Use `#human_attribute_name` instead

app/views/documents/edit.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        <div class="form-body">
+          <h2 class='text-center px-2 heading heading-2 strong-400 mb-0'>
+            <%= t '.document_edit' %>
+          </h2>
+          <%= form_with model: @document, local: true do |form| %>
+            <div class="col">
+              <div class="col-lg-12">
+              </br>
+                <div class="field">
+                  <%= form.label :category, t('activerecord.attributes.document.category')%><br />
```
Mate2xo: use `#human_attribute_name` instead

app/views/documents/_index.html.erb
```diff
@@ -21,8 +21,8 @@
               <table id="documents" class="table table-sm table-hover">
                 <thead>
                   <tr>
-                    <th scope="col">Nom du document</th>
-                    <th scope="col">Télécharger</th>
+                    <th scope="col"> <%= t 'activerecord.attributes.document.filename' %></th>
```
Mate2xo: Use `#human_attribute_name` instead

Mate2xo (changes requested):
I haven't finished yet, I'll make antother review later. Good work so far, thanks

app/views/documents/_index.html.erb
```diff
@@ -21,8 +21,8 @@
...
+                    <th scope="col"><%= t 'documents.download_file' %></th>
```
Mate2xo: You can use the I18n lookup shortcut here, by prefixing the string with a `.`
It's actually pretty useful to get some guidance on how to split our locale files and how to organize our locale keys.
For example, here if you do `t '.download_file'`, you'll have a 'missing translation' on the `documents.index.download_file` (guessing here). Doing so would make it clear that the locale keys we need here are only scoped to this view

app/views/documents/edit.html.erb
```diff
@@ -0,0 +1,44 @@
...
+                </div>
+                <br>
+
+              </div>
+            </div>
+            <br />
+
+
+            <div class="actions">
+              <%= form.submit t("main_app.views.application.buttons.edit_submit"), class: "btn-update btn btn-styled btn-lg btn-block btn-base-1 mt-4" %>
```
Mate2xo: How would `t('.edit_submit')` work here ?

config/locales/views/documents.en.yml
```diff
@@ -0,0 +1,6 @@
+en:
+  documents:
+    document_name: File name 
+    download_file: Download
```
Mate2xo: ref the the previous `.key.lookup` comment before ;
`download_file` sould probably be under the `index` key

Mate2xo (changes requested):

app/views/documents/edit.html.erb
Creiwry: regarding this comment, is it then best practice to add another key to views/documents so it's clear it's scoped to this view? .edit_submit does not work here.

Mate2xo: The `missing translation` error should actually guide you on wherei to put the required locale key.

e.g.
`translation missing : 'documents.edit.edit.submit` would mean you should put the locale as

```yml
fr:
  documents:
    edit:
      edit_submit: 'your string here'
```

Mate2xo: btw, since the locale scope would already be under an `edit` key, you could simplify the key name : 
```yml
fr:
  documents:
    edit:
      submit: 'your string'
```
which woud be called as
```ruby
<%= form.submit t(".edit"), class: "btn-update btn btn-styled btn-lg btn-block btn-base-1 mt-4" %>
```

app/controllers/documents_controller.rb
```diff
@@ -37,7 +52,11 @@ def destroy
   private
 
   def permitted_params
-    params.require(:document).permit(:category, :file)
+    params.require(:document).permit(:category, :file, :file_name)
+  end
+
+  def update_transaction # removed the equals thing here, don't know if it still works
```
Mate2xo, line 58: This comment does not gives relevant information to the reader without context. It's better to delete it.
Btw, if you don't know if it still works, then you should write a test to check that.

app/controllers/documents_controller.rb
```diff
@@ -22,6 +26,17 @@ def create
     end
   end
 
+  def update
+    @document = authorize Document.find(params[:id])
+    if update_transaction.success?
+      flash[:notice] = t 'activerecord.notices.messages.update_success'
```
Mate2xo: - Try the lookup `t '.success'`. You will then have an idea on where to put that key. The `activerecord` namespace isn't appropriate here, as we are not in an ActiveRecord object.
- On how to organize locale files, the easiest to maintain is to mimic the actual project structure. So that lookup key would ideally go in a file called `config/locales/controllers/documents.fr.yml`. Like that, we know exactly where are the locale file for a given object.

app/transactions/documents/update_transaction.rb
```diff
@@ -0,0 +1,42 @@
...
+    include Dry::Transaction
+
+    step :change_filename
+    step :update_file
+
+    def change_filename(params)
+      blob = get_blob(params[:document])
+      new_filename = "#{params[:file_name]}#{blob.filename.extension_with_delimiter}"
+
+      if params[:file_name].blank?
```
Mate2xo: You can actually make a guard clause for this conditional statement, as if there is no `params[:file_name]`, we don't need to process anything; so no need to `get_blob`, nor contructing the `new_filename`:
```ruby
def change_filename(params)
  return Success(params) if params[:file_name].blank?

  blob = get_blob(params[:document])
...
```

config/locales/views/documents.en.yml
```diff
@@ -0,0 +1,7 @@
+en:
+  documents:
+    edit:
+      document_edit: Edit your file
```
Mate2xo: `'documents.edit.document_edit'` is a bit redundant.
Try `documents.edit.title`, as it's actually a title

config/locales/views/documents.fr.yml
```diff
@@ -0,0 +1,7 @@
+fr:
+  documents:
+    edit:
+      submit: Sauvegarder les changements
+      document_edit: Modifier votre document
```
Mate2xo: -> `'documents.edit.title'`

Mate2xo (changes requested):

app/controllers/documents_controller.rb
Creiwry: Ooops, my bad. Made the comment for myself but forgot to delete

Mate2xo (approved):

PR #333, opened 2023-01-27
Add sorting functionality to document index view

Creiwry:
When index loads for the first time, documents are sorted based on filename. When user clicks on documents table header links, they can choose if they will sort based on filename or based on created_at. user can switch between ascending and descending order by clicking on header links.

Controller
- the default documents variable is @documents_by_name
- the other documents variable is @documents_by_date 

View
- added hidden renders of documents sorted by name asc/desc and date asc/desc
- added attributes to links in order to enable javascript functionality

documents.js
javascript functionality for the documents/_index partial
When the link Date asc gets clicked:
- link gets replaced with the other link in the pair: Date desc
- the documents in the order Date asc get rendered (the render of the partial containing these documents gets unhidden while all other documents partials get hidden)
- the other pair of links (Name asc and Name desc) get set to default: Name asc link is unhidden, Name desc link is hidden


app/views/documents/_index.html.erb
```diff
@@ -21,15 +21,16 @@
               <table id="documents" class="table table-sm table-hover">
                 <thead>
                   <tr>
-                    <th scope="col">Nom du document</th>
+                    <th scope="col"><%= link_to "Name", documents_path(sort: "filename", direction: "DESC", category: category.to_s), class:"name desc", id:"sort-name", remote: true %></th>
```
Mate2xo: The `#link_to` name (string)  should be an I18n lookup. Same for line 26.

app/views/documents/_index.html.erb
```diff
@@ -21,15 +21,16 @@
...
                     <th scope="col">Télécharger</th>
+                    <th scope="col"><%= link_to "Date", documents_path(sort: "created_at", direction: "ASC", category: category.to_s), class:"created_at", id:"sort-date", remote: true %></th>
                   </tr>
                 </thead>
 
                 <tbody>
                   <%= render partial: "documents/document", collection: @documents.select {|document|
                     document.category == category 
-                  } %>
+                  }, id: "documents" %>
```
Mate2xo: for method calls with multiline arguments, please align them together, and write one argument per line. This makes reading and reviewing easier, as we can see at a glance what are the arguments, and what is actually changed in a git diff

e.g.:
```ruby
<%= render partial: "documents/document",
           collection: @documents.select { |document| document.category == category },
           id: "documents" %>
```
-> we can see without reading that there are 3 arguments, and futures changes on this will be more obvious to review

app/views/documents/index.js.erb
```diff
@@ -0,0 +1,42 @@
+$("tbody").replaceWith("<tbody><%= j render partial: "documents/document", collection: @documents.select {|document|
```
Mate2xo: 1 argument per line if you make multiline method calls please.
You could also add new line after the first HTML tag, and before the last one, if that makes it easier to read

app/views/documents/index.js.erb
```diff
@@ -0,0 +1,42 @@
...
+      name_desc();
+    };
+    date_asc();
+  <% else %>
+    name_desc();
+  <% end %>
+}
+<%# I need to find a way to require .js.erb in this file so all I'll need to do is execute the function%>
+
+sort_class()
```
Mate2xo: Put this method call before all the function definitions, as this helps the reader understand what we actually do here, without reading the whole file first. A bit like a summary. JS functions are hoisted before the actual code is run, so you can call them from anywhere. So the ideal writing style would be to define methods by "abstraction level", which allows the reader to not having to read code that they don't care about.
e.g.
```javascript
sortClass();

function sortClass() { ... }
function dateDesc() { ... }
...
```

The naming of `#sort_class()` should be improved (and camelCased, it's the JS convention), because this function does not sort classes, it does not sort anything either. The method name should actually describe what is done, otherwise we force the reader to read everything to understand what is actually done.

app/views/documents/index.js.erb
```diff
@@ -0,0 +1,42 @@
+$("tbody").replaceWith("<tbody><%= j render partial: "documents/document", collection: @documents.select {|document|
+                    document.category == @category} %></tbody>");
+
+<%# I'd like to add the code between the comments to a different file, so that this file is organized better %>
```
Mate2xo: I thing you should be able to `import` any JS file here. But you might need to add it to the asset pipeline first.
Try to do that, but if it takes you more than 30mins, we'll do that together

app/controllers/documents_controller.rb
```diff
@@ -3,11 +3,19 @@
 # Document management
 class DocumentsController < ApplicationController
   def index
+    @sorting_params = [params[:sort], params[:direction]]
+    @category = params[:category]
+    join_sql = "INNER JOIN active_storage_attachments ON active_storage_attachments.record_id = documents.id INNER JOIN active_storage_blobs ON active_storage_blobs.id = active_storage_attachments.blob_id"
```
Mate2xo: Do we really need such a join ? I think we might not, since the `Document` record seems to have all the info we need (`:created_at` and `:filename`)
If we do need it, is there a way to write it without an SQL string, maybe with options in the `#joins` method ?
And if not, the `join_sql` variable should have a better name. We can see that's SQL, it would be better if the variable name actually reflects 'what' it does, or 'why' it does it

app/controllers/documents_controller.rb
```diff
@@ -3,11 +3,19 @@
...
+
+    respond_to do |format|
+      format.js
+      format.html
+    end
```
Mate2xo: Since you don't pass any options to the `.js` and `.html` statements, I think you can omit this block entirely. The controller will by default render the action-matching `.js` template when it receives a JS request

Mate2xo (changes requested):

app/views/documents/index.js.erb
Creiwry: Hi Tim, I followed your advice on the comment above, but I can't in this case as jQuery is white space sensitive.

app/assets/javascripts/documents.js
```diff
@@ -0,0 +1,46 @@
+console.log("ready");
```
Mate2xo: A wild console.log has appeared !

app/assets/javascripts/documents.js
```diff
@@ -0,0 +1,46 @@
...
+  if(this.classList.contains("name")){
+    document.getElementById(currentCategory.split("_").join("-") + "-date-asc").hidden =false;
+    document.getElementById(currentCategory.split("_").join("-") + "-date-desc").hidden =true;
+  } else {
+    document.getElementById(currentCategory.split("_").join("-") + "-name-asc").hidden =false;
+    document.getElementById(currentCategory.split("_").join("-") + "-name-desc").hidden =true;
+  };
+};
+
+sortContent()
```
Mate2xo: Put this at the top of the file, and the functions at the bottom. JS hoisting will make your defined function available at runtime

app/assets/javascripts/documents.js
```diff
@@ -0,0 +1,46 @@
+console.log("ready");
+function sortContent(){
```
Mate2xo: This doesn't sort, it inits the sorting functionnality. So rename it to `initDocumentsSorting`
Also, since this is for the documents index, rename the file to `documents/index.js`

app/assets/javascripts/documents.js
```diff
@@ -0,0 +1,46 @@
...
+  let links = [...this.parentNode.childNodes];
+
+  links[1].hidden = !links[1].hidden || false;
+  links[3].hidden = !links[3].hidden || false;
+};
+
+function chooseContentToDisplay(){
+  let currentLinkId = this.id;
+  let currentCategory = this.dataset['category'];
+  let selector = '.sort-pane[data-category="' + currentCategory + '"]';
```
Mate2xo: Use String interpolation

app/assets/javascripts/documents.js
```diff
@@ -0,0 +1,46 @@
...
+
+  for (let i = 0; i  < sortingLinks.length; i ++) {
+    sortingLinks[i].addEventListener("click", switchLinkBetweenAscendingAndDescending);
+    sortingLinks[i].addEventListener("click", chooseContentToDisplay);
+    sortingLinks[i].addEventListener("click", resetOtherLink);
+  };
+};
+
+function switchLinkBetweenAscendingAndDescending() {
+  let links = [...this.parentNode.childNodes];
```
Mate2xo: wouldn't 
```javascript
let links = this.parentNode.childNodes;
```
give exactly the same result ?

app/assets/javascripts/documents.js
```diff
@@ -0,0 +1,46 @@
...
+
+  links[1].hidden = !links[1].hidden || false;
+  links[3].hidden = !links[3].hidden || false;
+};
+
+function chooseContentToDisplay(){
+  let currentLinkId = this.id;
+  let currentCategory = this.dataset['category'];
+  let selector = '.sort-pane[data-category="' + currentCategory + '"]';
+  let content = [...document.querySelectorAll(selector)];
```
Mate2xo: naming suggestion
`content` -> `contentToDisplay` ?

app/assets/javascripts/documents.js
```diff
@@ -0,0 +1,46 @@
...
+      content[i].hidden = true;
+    };
+  };
+};
+
+function resetOtherLink(){
+  let currentCategory = this.dataset['category'];
+
+  if(this.classList.contains("name")){
+    document.getElementById(currentCategory.split("_").join("-") + "-date-asc").hidden =false;
```
Mate2xo: unecessary snake_case to kebab-case conversion. Just use the actual `id` attribute value, and String interpolation.
And add a space after the `=` signs

app/assets/javascripts/documents.js
```diff
@@ -0,0 +1,46 @@
...
+  for (let i = 0; i < content.length; i ++) {
+    if(content[i].classList.contains(currentLinkId)){
+      content[i].hidden = false;
+    } else {
+      content[i].hidden = true;
+    };
+  };
+};
+
+function resetOtherLink(){
```
Mate2xo: `resetOtherLink` -> `resetOtherSortingLink`

Mate2xo (changes requested):

app/controllers/documents_controller.rb
```diff
@@ -4,11 +4,28 @@
 class DocumentsController < ApplicationController
   def index
     @document = Document.new
```
Mate2xo, line 6: There are common query setup for each variable values that you can dry up.
e.g.
```ruby
base_relation  = Document.with_attached_file
base_relation = base_relation.where(published: true) if member_signed_in?
...
```

app/controllers/documents_controller.rb
```diff
@@ -35,7 +52,6 @@ def destroy
   end
 
   private
-
```
Mate2xo: keep the empty line after `private`

app/views/documents/_document.html.erb
```diff
@@ -1,24 +1,28 @@
...
+                      rails_blob_path(document.file, disposition: "attachment") %>
+        </td>
+        <td scope="row">
+          <% if document.file.previewable? %>
+            <%#= link_to(image_tag(document.file.preview(resize: "90x90")),
+                        rails_blob_path(document.file, disposition: "attachment")) %>
+          <% else %>
+            <%#= link_to document.file.filename,
+                rails_blob_path(document.file, disposition: "attachment") %>
+          <% end %>
```
Mate2xo: `#`

And there's apparently an indentation problem

app/models/document.rb
```diff
@@ -37,4 +37,12 @@ class Document < ApplicationRecord
...
     'text/plain'
   ]
+  def self.join_documents_to_filenames
+    joins(<<-SQL)
+      INNER JOIN active_storage_attachments
+      ON active_storage_attachments.record_id = documents.id
+      INNER JOIN active_storage_blobs
+      ON active_storage_blobs.id = active_storage_attachments.blob_id
+    SQL
+  end
```
Mate2xo: * Use `#scope` to set up this (check the rails docs)
* I think you can use `#joins` here instead of manually building the SQL string. E.g.: `joins(active_storage_attachments: [:active_storage_blobs])`
* this join might not be necessary. Try something like this : https://medium.com/fiatinsight/how-to-change-a-filename-in-rails-active-storage-f3e4f26f427e
* -> `join_filenames_to_documents`

app/views/documents/_index.html.erb
```diff
@@ -16,20 +17,64 @@
...
-                  <tr>
-                    <th scope="col">Nom du document</th>
-                    <th scope="col">Télécharger</th>
+                  <tr >
+                    <th scope="col">
+                      <li style="list-style: none;">
+                        <a class="sort-link name" 
+                           data-category="<%= category %>"
+                           href="#"
+                           data-remote="true"
```
Mate2xo: `data-remote: true` should not be needed here, as you are not making any backend request

app/views/documents/_index.html.erb
```diff
@@ -16,20 +17,64 @@
...
+                           id="<%= category.split("_").join("-") %>-name-desc"><%= t ".name" %></a>
```
Mate2xo: Unnecessary string transformation

config/initializers/assets.rb
```diff
@@ -17,7 +17,7 @@
 # folder are already added.
 Rails.application.config.assets.precompile += %w(
   fullCalendar.js fullCalendar.css
-  openStreetMaps.js
+  openStreetMaps.js documents.js
```
Mate2xo: put the new js file declaration on its own line (every file here are kinda classified by their scope

Mate2xo (changes requested):

Issue #338, opened 2023-02-26
Proficiency check on slot validation


Mate2xo:
### TLDR
Proficiency validation doesn't seem to be active on a Mission slot
We cannot update a member's slot if this slot is full

### As a 
User, or admin user

### When I'm
on the Mission creation/edit form

### Current behaviour
I can fill in the whole Mission slot with members that are not proficient on the cash register
I cannot update a member's time slot if the mission slot is full

### Desired behaviour
* The Mission slot should not be valid if there is not at least one assigned member that is proficient on the cash register
* We can update a member's time slot even if the mission slot is full
* Add a new proficiency level : `newbie` (`nouveau` in french)

### Additional infos
Maybe this only happens for admin users ? This validation was supposed to exist

PR #339, opened 2023-02-27
Fix validation bug on missions update #338

Creiwry:
Issue:
The enrollments validator was raising an AvailabilityPlaceValidator error when enrollments were updated. Only this issue from the ticket is being addressed here due to priority.

Implemented changes:
The AvailabilityPlaceValidator  scope has been changed to on: create, as that is the only time it is helpful to check if the max_member_count condition has been reached.

### DIFF
--- a/spec/models/enrollment_spec.rb
+++ b/spec/models/enrollment_spec.rb
@@ -25,6 +25,18 @@ RSpec.describe Enrollment, type: :model do
     end
   end
 
+  describe 'validations' do
+    context 'when a new enrollment is created where mission is full' do
+      it 'raises an error' do
+        mission =  create :mission, max_member_count: 4 
+        enroll_n_members_on_mission(mission, 4)
+        extra_enrollment = Enrollment.new(mission: mission)
+
+        expect(extra_enrollment).to_not be_valid
+      end
+    end
+  end
+
   describe '#duration' do
     subject(:create_enrollment) do
       create :enrollment, start_time: mission.start_date, end_time: mission.start_date + 2.hours, mission: mission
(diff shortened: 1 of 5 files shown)


Mate2xo:
@Creiwry To be sure we get the intended result, please write a test to simulate what happens on a full slot when another member tries to join in

spec/models/enrollment_spec.rb
```diff
@@ -25,6 +25,18 @@
     end
   end
 
+  describe 'validations' do
+    context 'when a new enrollment is created where mission is full' do
```
Mate2xo: [...] is created `on a full mission`
might be better ?

spec/models/enrollment_spec.rb
```diff
@@ -25,6 +25,18 @@
...
+      it 'raises an error' do
+        mission =  create :mission, max_member_count: 4 
+        enroll_n_members_on_mission(mission, 4)
+        extra_enrollment = Enrollment.new(mission: mission)
+
+        expect(extra_enrollment).to_not be_valid
+      end
```
Mate2xo: Instead of creating a method helper that is used only once in that file (meaning, it only moves the logic elsewhere), I think this setup can be done directly under this context. You could either create a new `trait` in the `mission` factory, or directly set the setup here in a `let`. Here, the `let` directly illustrates what is the actual context in terms of code

```suggestion
      let(:full_mission) do
        create :mission, max_member_count: 4 do |mission|
            create_list :enrollment,
                         4,
                         mission: mission,
                         start_time: mission.start_time,
                         end_time: mission.start_date + 3.hours
          end
        end
      end
      
      it 'is invalid' do
        extra_enrollment = Enrollment.new(mission: full_mission)

        expect(extra_enrollment).to_not be_valid
      end
```

Also, it's better to be specific if we can on what error is raised. Here, it's an invalidation

Mate2xo (changes requested):

PR #348, opened 2025-05-20
Errors reporting

Mate2xo:
* integration  with Appsignal
* set distinct config for staging env
* improve system test perf
* promoted test helper to all system tests, as it's needed everywhere to use the headless mode


spec/support/helpers/system_browser.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def use_fast_non_js_browser
+    driven_by(:rack_test)
+  end
+
+  def wait_for_flash_messages
+    expect(page).to have_css('div#flash_messages')
+  end
+
+  def wait_flash(color)
```
Floury-Fabio: suggestion: 
the name `wait_flash` doesn't seem follow the ecosystem rails conventions
We can rename this method wait_for_alert or expect_alert

We can also use wait_for_flash_messages for this method and remove the old method. The old method is not used. It was a fallback method that allowed the dev to wait for the flash without making any assumptions about its type

spec/system/document_uploads_at_document_page_section_spec.rb
```diff
@@ -47,25 +49,24 @@
...
-
-    before do
+      use_headless_javascript_browser
       sign_in create(:member, :admin)
       create(:document)
       visit documents_path(anchor: 'documents')
+
+      click_on I18n.t('main_app.views.application.buttons.destroy')
+      page.driver.browser.switch_to.alert.accept
     end
```
Floury-Fabio: I agree with for the remove of the before, but  solution mixes several subjects and forward the problem of the before in the subjet.

my problems with this subject: 
-  the subject shouldn't contain the helper config `use_headless_javascript_browser`, this helper is not in the scenario from the point of view of the user.
- The subject merge the preparation and the exercice phase.
- The SUT is no longer explicit

proposition:

```
context 'context' do
    before { use_headless_javascript_browser }

    it 'it' do
      sign_in create(:member, :admin)
      create(:document)

      visit documents_path(anchor: 'documents')
      click_on I18n.t('main_app.views.application.buttons.destroy')
      page.driver.browser.switch_to.alert.accept

      expect(page).to have_no_content 'erd.pdf'
    end
  end
```
the before is still there but it is reserved for configuration and can be easily ignored in the context of the test.
The sut and the phases are explicit.
Repetition across tests is not a problem. Dry is counterproductive in testing.

The doc that guided me towards this proposal:

- https://thoughtbot.com/blog/the-self-contained-test
- https://thoughtbot.com/blog/how-we-test-rails-applications
- https://thoughtbot.com/blog/a-journey-towards-better-testing-practices
- https://github.com/thoughtbot/guides/tree/main/testing-rspec
- https://github.com/thoughtbot/guides/blob/main/testing-rspec/avoid_let_spec.rb

README.md
```diff
@@ -54,10 +54,17 @@ Members buy products from productors, and sell these products to themselves. Eac
...
-- Ruby => 2.5.1
-- Rails => 5.2.3
+Error monitoring is managed through [Appsignal](https://www.appsignal.com/) who graciously offered us a sponsored account, through their [open source initiative](https://www.appsignal.com/open-source), to allow us to track error and performance issues. Thank you Roy!
+I personally recommend their app for their reliability and ease of use, as I've used it already for around 40 other production Rails apps :thumbsup:
+
+If another, or none, error reporting service is needed, we could add a `Monitoring` adapter to offer configuration options for this feature.
+
+## Versions
+
+- Ruby => 2.7.8
```
Floury-Fabio: you can change the versions :)

spec/support/helpers/system_browser.rb
Mate2xo: What conventions are you referring to ? You mean the default flash types ? (`alert` & `notice`)

Mate2xo: Since the behaviour was previously exactly the same in these methods (except from the flash type)
```ruby
  def wait_for_success_flash
    expect(page).to have_css('div#flash_notice.alert-success')
  end

  def wait_for_warning_flash
    expect(page).to have_css('div#flash_alert.alert-warning')
  end

  def wait_for_danger_flash
    expect(page).to have_css('div#flash_error.alert-danger')
  end
```
I thought it was easier to read if we parameterized these types

Mate2xo: Since they are actual expectations, I promoted them to be Matchers

spec/system/member_count_limit_on_missions_spec.rb
```diff
@@ -1,66 +1,71 @@
...
 
-        visit mission_path(mission.id)
-        click_button I18n.t('main_app.views.missions.show.button_enroll')
+    context 'when the enrolled Member count has been reached' do
+      let(:mission) do
+        create(:mission) do |mission|
+          mission.max_member_count = 4
+          mission.members << create_list(:member, 4)
+          mission.save
+        end
```
Floury-Fabio: gray zone: you can avoid a save with factorybot params

spec/system/members/cash_register_proficiency_spec.rb
```diff
@@ -6,7 +6,10 @@
   let(:mission) { create(:mission) }
   let(:jack) { create(:member, cash_register_proficiency: :proficient) }
 
-  before { sign_in jack }
+  before do
```
Floury-Fabio: gray zone: you shadow the driver config. you can instead make a config driver for each context putting the driver config in the before L15

Floury-Fabio (dismissed):

spec/system/members/cash_register_proficiency_spec.rb
```diff
@@ -24,13 +27,14 @@
 
   context 'when on the mission index page' do
     before do
+      use_headless_javascript_browser
       mission.members << create_list(:member, 3, cash_register_proficiency: :untrained)
       visit missions_path
     end
 
     it 'shows missions without proficient members in purple', :js do
```
Floury-Fabio: with driver config the js option is not necessary

Floury-Fabio (approved):

PR #350, opened 2025-05-30
Rails7 upgrade

Mate2xo:
Upgrade from Rails 6.0 to 7.1, with fixes along the way.
* missing translations, and extractions to narrower namespaces
* update LSP config
* improve coverage
* ActiveAdmin had to be upgraded from 2 -> 3 for Rails 7.1; which needed upgrading Ransack from 3 -> 4, that requires whitelisting searchable attributes. Admin filters have been improved and simplified along the way
* bumped other gems that would'nt work with Rails 7.0/7.1

There's a lot of lint, this was done on separate commits to ease the review. It might be easier to review this PR commit by commit


db/schema.rb
```diff
@@ -22,19 +21,19 @@
     t.bigint "resource_id"
     t.string "author_type"
     t.bigint "author_id"
-    t.datetime "created_at", precision: 6, null: false
-    t.datetime "updated_at", precision: 6, null: false
-    t.index ["author_type", "author_id"], name: "index_active_admin_comments_on_author_type_and_author_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["author_type", "author_id"], name: "index_active_admin_comments_on_author"
```
Floury-Fabio, line 26: care the `db:migrate` command generate this diff 
Screenshot from 2025-05-31 17-25-45

is the name change intentional?

look up the rails version specified in schema.

Mate2xo: Yes, I started all migrations from a fresh DB just to be sure.
Although doing that revealed that a few more changes were needed (explained in the commit messages)

Floury-Fabio (approved):

